PR #12, opened 2016-09-20
Use expressions in Filter and Project

ajnavarro:
- Fixed type boolean checks


sql/expression/boolean.go
```diff
@@ -4,14 +4,16 @@ import "github.com/mvader/gitql/sql"
 
 type Not struct {
 	child sql.Expression
+	name  string
```
mvader: what's the name for?

PR #21, opened 2016-09-20
Implement Node in Limit

dpordomingo:


sql/plan/limit.go
```diff
@@ -22,6 +22,11 @@ func (l *Limit) Schema() sql.Schema {
 	return l.UnaryNode.Child.Schema()
 }
 
+func (l *Limit) Resolved() bool {
+	//TODO: what "Resolvable" interface do?
```
mvader: it tells if the node is resolved or not (needs further changes during the analysis part)

PR #36, opened 2016-11-04
git: Add tags relation

ajnavarro:


git/tags.go
```diff
@@ -0,0 +1,85 @@
...
+	return []sql.Node{}
+}
+
+type tagIter struct {
+	i *git.TagIter
+}
+
+func (i *tagIter) Next() (sql.Row, error) {
+	tag, err := i.i.Next()
+	if err == io.EOF {
```
smola: This block is equivalent to the next one.

git/tags_test.go
```diff
@@ -0,0 +1,47 @@
+package git
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/fixtures"
+	"time"
```
smola: "time" in the std block.

PR #84, opened 2016-12-10
git, sql: create a TableBase to avoid code duplication

ajnavarro:


sql/table_base.go
```diff
@@ -0,0 +1,21 @@
+package sql
+
+type TableBase struct {
+	Table
```
smola: this Table is not needed, so `type TableBase struct{}` should be enough.

Issue #82, opened 2016-12-02
cli: add flag to output csv results


smola:
In order to use gitql for scripting, CSV output is more convenient than formatted tables.

Issue #91, opened 2016-12-31
SQLite3 database export?


EtherTyper:
I think being able to export a db dump of all the [tables described in README.md](https://github.com/gitql/gitql#tables) would be very useful.

smola:
@EtherTyper CSV output will come soon, see https://github.com/gitql/gitql/issues/82
I think that should be enough to export to other databases.

Anyway, we might accept contributions of functionality to get the output as a SQL dump.

smola:
It is now possible to query tables with `SELECT * FROM <table>` and dump to a CSV or JSON file.
Closing this, since this is probably as far as we want to go for export. We might add a `dump` command to dump full databases at some point.

EtherTyper:
@smola Cool! I'll try that out.

PR #93, opened 2017-01-13
cmd: added new output formats

ajnavarro:
Now you can change the output format to 2 new ones: json and csv.

- Added json support
- Added csv support
- Now the table output format is called 'pretty'. It is the default format.

### DIFF
--- a/internal/format/csv.go
+++ b/internal/format/csv.go
@@ -1,0 +1,36 @@
+package format
+
+import (
+	"encoding/csv"
+	"fmt"
+	"io"
+)
+
+type CsvFormat struct {
+	cw *csv.Writer
+}
+
+func NewCsvFormat(w io.Writer) *CsvFormat {
+	return &CsvFormat{
+		cw: csv.NewWriter(w),
+	}
+}
+
+func (cf *CsvFormat) WriteHeader(headers []string) error {
+	return cf.cw.Write(headers)
+}
+
+func (cf *CsvFormat) Write(line []interface{}) error {
+	rowStrings := []string{}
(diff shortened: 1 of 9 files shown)


ajnavarro:
It Closes #82 and is related to #91

internal/format/csv.go
```diff
@@ -0,0 +1,36 @@
...
+	cw *csv.Writer
+}
+
+func NewCsvFormat(w io.Writer) *CsvFormat {
+	return &CsvFormat{
+		cw: csv.NewWriter(w),
+	}
+}
+
+func (cf *CsvFormat) Init(headers []string) error {
```
smola: These method names could look more standard:
Init -> WriteHeader
Line -> Write
End -> Close

internal/format/json.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (cf *JsonFormat) Init(headers []string) error {
+	cf.keys = headers
+
+	return nil
+}
+
+func (cf *JsonFormat) Line(line []interface{}) error {
+
```
smola: extra line

internal/format/format_test.go
```diff
@@ -0,0 +1,89 @@
...
+		Headers: []string{"a", "b", "c"},
+		Lines: [][]interface{}{
+			[]interface{}{
+				"a1", "b1", "c1",
+			},
+		},
+	},
+}
+
+func TestFormats(t *testing.T) {
```
smola: use separate functions for each format so that we have separate test output for them

PR #126, opened 2017-09-30
Fix import paths

kenshaw:
The import paths used in gitquery prevent it from being compiled
properly, due to inclusion of the internal path. Additionally, this was
likely not noticed since the cmd/gitquery/*.go files were making use of
the gopkg.in namespace. This changeset fixes those issues.


kenshaw:
Please note this has errored on Travis due to the go-git API change. If my previous PR #126 is included, this should build correctly.

kenshaw:
Please note the reason I am adding these changes, is that I'd like to add `gitquery` to the supported databases that the `usql` cli tool supports. That's not possible until the API + import paths are fixed.

ajnavarro:
Hi @kenshaw , Thanks a lot for your contributions.

The `gopkg.in/sqle/gitquery.v0` import is correct, you should use it to import gitquery in other projects. Also, to be able to compile it locally, you should develop inside the folder `$GOPATH/gopkg.in/sqle/gitquery.v0` instead of `$GOPATH/github.com/sqle/gitquery`. Doing this all the compilation errors will disappear.

Another question, why did you move all the files on the `internal` folder outside? Do you need them directly in a third party project?

kenshaw:
Are you using non-standard paths and/or a quite old version of Go?  I'm not sure if this is because of the weird gopkg.in organization or whatever:
```sh
ken@ken-desktop:~/src/go/src/github.com/sqle/gitquery/cmd/gitquery$ go build
query_base.go:9:2: use of internal package not allowed
```

I would highly suggest not using the `gopkg.in` for versioning this package -- if this is a dependency a project needs to rely on, then the using project should vendor the dependencies.

smola:
@kenshaw Just checkout the project into `$GOPATH/src/gopkg.in/sqle/gitquery.v0`, that is the intended usage at the moment. We might reconsider this before v1 though.

PR #143, opened 2018-02-27
Add server command, delete query and version

### LINKED ISSUE #140: Add server functionality
Add a server command to start a compatible mysql server.

jfontan:
Fixes #140


cmd/gitquery/server.go
```diff
@@ -0,0 +1,65 @@
...
+	if err := c.buildDatabase(); err != nil {
+		return err
+	}
+
+	auth := mysql.NewAuthServerStatic()
+	auth.Entries[c.User] = []*mysql.AuthServerStaticEntry{
+		{Password: c.Password},
+	}
+
+	hostString := fmt.Sprintf("%v:%v", c.Host, c.Port)
```
mcarmonaa: Do you think is valuable use `net.JoinHostPort` instead of `fmt.Sprintf`?

jfontan: Thanks for the tip. JoinHostPort supports IPv6 addresses so it's better.

jfontan:
Rebased to master

PR #146, opened 2018-03-06
[WIP] integration test with queries from QuerySetApp

mcarmonaa:
This test takes those queries from `QuerySetApp` that don't use UDFs. Apart of the expressions already implemented they need `CASE` and `HAVING` so we should implement them if you don't know other way to formulate the queries.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -1,0 +1,45 @@
+package test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/src-d/gitquery"
+	"github.com/src-d/gitquery/internal/function"
+	"github.com/stretchr/testify/require"
+	sqle "gopkg.in/src-d/go-mysql-server.v0"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+func TestIntegration(t *testing.T) {
+	engine := sqle.New()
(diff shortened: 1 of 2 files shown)


erizocosmico:
I guess until HAVING and CASE are implemented those queries are better left out and it's ok to test with what's available.

test/integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+				checkSchema(t, schema, query.expectedSchema)
+				checkIter(t, rowIter, query.expectedRows)
+			}
+		})
+	}
+}
+
+func checkSchema(t *testing.T, schema, expected sql.Schema) {
+	// TODO: add expected schema to the queries and check it
+	require.Nil(t, expected)
```
ajnavarro: change to `require.Equals(expected,schema)` to make it fails until the queries will be implemented

test/integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func checkIter(t *testing.T, rowIter sql.RowIter, expected int) {
+	// TODO: add expected number of rows to the queries and check it
+	require.Zero(t, expected)
```
ajnavarro: change to `require.Equals(expected,count)` to make it fails until the queries will be implemented

PR #149, opened 2018-03-07
gitquery: implement session

erizocosmico:
Depends on https://github.com/src-d/go-mysql-server/issues/36 and https://github.com/src-d/go-mysql-server/pull/87 (that's why CI fails)

### DIFF
--- a/internal/function/is_remote.go
+++ b/internal/function/is_remote.go
@@ -18,11 +18,9 @@ func NewIsRemote(e sql.Expression) sql.Expression {
 	return &IsRemote{expression.UnaryExpression{Child: e}}
 }
 
-var _ sql.Expression = (*IsRemote)(nil)
-
 // Eval implements the expression interface.
-func (f *IsRemote) Eval(row sql.Row) (interface{}, error) {
-	val, err := f.Child.Eval(row)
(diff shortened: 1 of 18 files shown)


internal/function/is_remote.go
```diff
@@ -18,11 +18,9 @@ func NewIsRemote(e sql.Expression) sql.Expression {
 	return &IsRemote{expression.UnaryExpression{Child: e}}
 }
 
-var _ sql.Expression = (*IsRemote)(nil)
```
mcarmonaa: Why now isn't needed to check that it implements the `sql.Expression` interface?

internal/function/is_tag.go
```diff
@@ -18,11 +18,9 @@ func NewIsTag(e sql.Expression) sql.Expression {
 	return &IsTag{expression.UnaryExpression{Child: e}}
 }
 
-var _ sql.Expression = (*IsTag)(nil)
```
mcarmonaa: Why now isn't needed to check that it implements the `sql.Expression` interface?

internal/function/is_remote.go
erizocosmico: `TransformUp` already makes this check

ajnavarro:
should we merge this?

erizocosmico:
@ajnavarro once I rebase, yes

PR #163, opened 2018-03-12
Set Source in table schemas

jfontan:


repositories.go
```diff
@@ -24,7 +24,7 @@ func (repositoriesTable) Name() string {
 
 func (repositoriesTable) Schema() sql.Schema {
 	return sql.Schema{
-		{Name: "id", Type: sql.Text, Nullable: false},
+		{Name: "id", Type: sql.Text, Nullable: false, Source: "repositories"},
```
mcarmonaa: Isn't there a constant for the repositories table name too?

jfontan: Changed the string to a constant and added test to repository table.

PR #164, opened 2018-03-12
gitquery: implement filter pushdown for git tables 

### LINKED ISSUE #153: implement filter and column pushdown
We should implement filter and column pushdown once they land in go-mysql-server.

I don't think we have any computationally expensive columns yet, so we shouldn't care much about column pushdown, but we should about filter pushdown, which can reduce a lot the amount of data sent.

erizocosmico:
Closes #153 

This PR implements filter pushdown for all git tables, drastically reducing the amount of data we push to the joins by applying some of the filters directly inside the tables.

The filters are split in two categories:

- Regular filters: any expression that references any field in the table that requires the concrete value of the row to be executed (e.g. comparison operators, negation, udfs, ...)
- Selectors: filters that are of the form `FIELD = LITERAL`, `FIELD = LITERAL1 OR FIELD = LITERAL2 OR FIELD = LITERALN` or `FIELD IN (LIT1, LIT2, ...)` (when it's implemented) that give us a set of concrete values to directly access those rows instead of iterating over all the data. Example: `SELECT * FROM commits WHERE hash = 'foo'` we'd only go and get the commit with the hash `foo` instead of iterating all of them and check if the hash matches.

Since all this logic is common to all the tables, I abstracted this logic into two functions:

- `handledFilters` this is basically the same for all of them, except for the schema and the table name, so the `HandledFilters` methods just calls `handledFilters` function with some params based on the table.
- `rowIterWithSelectors` this is perhaps too much of an abstraction, but avoids putting so much boilerplate inside all the `WithProjectAndFilters` method, which is exactly the same. Basically it needs all the schema, session, table name, filters that will be used as selectors, etc and then a callback that given a set of selectors will return a `RowRepoIter` and the rest is handled behind the scenes.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -1,14 +1,24 @@
 package gitquery
 
 import (
+	"io"
+
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 
+	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
 )
 
 type blobsTable struct {
 	pool *RepositoryPool
 }
+
+var blobsSchema = sql.Schema{
+	{Name: "hash", Type: sql.Text, Nullable: false, Source: blobsTableName},
+	{Name: "size", Type: sql.Int64, Nullable: false, Source: blobsTableName},
+}
+
+var _ sql.PushdownProjectionAndFiltersTable = (*blobsTable)(nil)
 
 func newBlobsTable(pool *RepositoryPool) sql.Table {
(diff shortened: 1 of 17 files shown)


erizocosmico:
Fixed. Can you guys please provide some feedback on the selectors thing? It maybe be a bit of an overcomplicated (even a bit counterintuitive, even if it's thoroughly documented) concept for this, but that's the most effective way of handling this I could think of.

jfontan:
It does not seem that counterintuitive. Also these are implemented once per table. Even if the tables are changed this logic should stay the same.

PR #165, opened 2018-03-13
gitquery: add integration tests skeleton

erizocosmico:


jfontan:
There is already a base integration test here https://github.com/src-d/gitquery/pull/146. We should mix both or use this one with the queries from the pull request.

ajnavarro:
Yep, but is not checking the output. We should check it, as in this PR.

erizocosmico:
Moving this to #166, since this is empty and the other contains one real test plus changes to the skeleton.

PR #166, opened 2018-03-13
 it: blobs per master commit 

### LINKED ISSUE #156: Add integration test to get the number of blobs per commit from HEAD references

erizocosmico:
Closes #156 

Just with master instead of HEAD. HEAD ref is not listed, have to check why, but for this purpose it doesn't matter which branch it is.


integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	_, err := pool.AddGit(path)
+	require.NoError(t, err)
+
+	engine.AddDatabase(gitquery.NewDatabase("foo", &pool))
+	function.Register(engine.Catalog)
+
+	testCases := []struct {
+		query    string
+		result   []sql.Row
+		sortFunc rowSortFunc
```
erizocosmico: Some queries, such as the aggregations can't be sorted using `ORDER BY`, so we need this to check some outputs :(

ajnavarro: Maybe you can use `require.ElementsMatch` instead of `require.Equal` to check the output

erizocosmico:
Fixed @ajnavarro

PR #167, opened 2018-03-13
 gitquery: return HEAD reference on refs table

erizocosmico:
Depends on #166 

HEAD reference was not being returned in the references table.

### DIFF
--- a/references.go
+++ b/references.go
@@ -55,6 +55,7 @@ func (referencesTable) Children() []sql.Node {
 }
 
 type referenceIter struct {
+	head         *plumbing.Reference
 	repositoryID string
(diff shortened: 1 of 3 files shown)


references.go
```diff
@@ -65,7 +66,13 @@ func (i *referenceIter) NewIterator(repo *Repository) (RowRepoIter, error) {
 		return nil, err
 	}
 
+	head, err := repo.Repo.Head()
```
jfontan: In rooted repos head is not set. Can this cause problems?

jfontan: On a second though only normal git repos are supported right now.

PR #169, opened 2018-03-13
internal/function: add commit_has_tree UDF

### LINKED ISSUE #152: commit_has_tree(commit_hash, tree_hash) UDF
We have ways of joining almost all tables in gitquery, except commits and trees.

- repositories and remotes join by repo id.
- repositories and refs join by repo id.
- refs and commits join using history_idx udf.
- commits and blobs join using commit_contains udf.
- trees and blobs join using blob.hash and tree_entries.entry_hash.
- trees and commits don't have anything to join.

We could have another UDF to perform the join between these two tables:

```
commit_has_tree(commit_hash, tree_hash)
```

For example, select commit messages with Go files:

```sql
SELECT message 
FROM commits 
  INNER JOIN tree_entries 
  ON commit_has_tree(hash, tree_hash) 
WHERE name LIKE '%.go';
```

If we add this UDF, I also propose to rename `commit_contains` to `commit_has_blob`, because commit can contain many things and is consistent with this naming.

Thoughts? /cc @mcarmonaa @jfontan @ajnavarro

erizocosmico:
Closes #152 

# Important pls readme

Note that another, perhaps simpler, solution would be to add a `tree` column to the `commits` table (since the tree hash is in the commit object) and then just `commits.tree = tree_entries.tree_hash`

### DIFF
--- a/internal/function/commit_has_tree.go
+++ b/internal/function/commit_has_tree.go
@@ -1,0 +1,171 @@
+package function
+
+import (
+	"io"
+
+	"gopkg.in/src-d/go-git.v4/plumbing/filemode"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+
+	"github.com/src-d/gitquery"
+	git "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
+)
+
+// CommitHasTree is a function that checks whether a tree is part of a commit
+// or not.
+type CommitHasTree struct {
+	expression.BinaryExpression
+}
+
+// NewCommitHasTree creates a new CommitHasTree function.
+func NewCommitHasTree(commit, tree sql.Expression) sql.Expression {
+	return &CommitHasTree{expression.BinaryExpression{
+		Left:  commit,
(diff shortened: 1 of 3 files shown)


ajnavarro:
> Note that another, perhaps simpler, solution would be to add a tree column to the commits table (since the tree hash is in the commit object) and then just commits.tree = tree_entries.tree_hash

We should add also the `tree_hash` column to `commits` table, but what happens if you want to check if a tree that is not the parent one is on that commit?

internal/function/commit_has_tree.go
```diff
@@ -0,0 +1,120 @@
...
+func (f *CommitHasTree) commitHasTree(
+	repo *git.Repository,
+	commitHash, treeHash plumbing.Hash,
+) (bool, error) {
+	commit, err := repo.CommitObject(commitHash)
+	if err != nil {
+		return false, err
+	}
+
+	return commit.TreeHash == treeHash, nil
```
ajnavarro: We should check all the tree objects, not only the root one.

erizocosmico: Only other thing related to trees `*object.Commit` has is `Tree`, which returns the tree pointed to by `TreeHash`, and then the only thing this has is the `TreeEntries`. Do you mean that? Wouldn't that be a tree entry and not a tree?

erizocosmico:
Fixed @ajnavarro, now it checks in all trees.

PR #175, opened 2018-03-14
gitquery: add content column to blobs table

### LINKED ISSUE #172: Add content column to blob table

erizocosmico:
Closes #172

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -1,6 +1,7 @@
 package gitquery
 
 import (
+	"bufio"
 	"io"
 	"io/ioutil"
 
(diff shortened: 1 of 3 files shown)


blobs.go
```diff
@@ -142,17 +144,28 @@ func (i *blobsByHashIter) Next() (sql.Row, error) {
...
 }
 
-func blobToRow(c *object.Blob) sql.Row {
+func blobToRow(c *object.Blob) (sql.Row, error) {
+	r, err := c.Reader()
+	if err != nil {
+		return nil, err
+	}
+
+	content, err := ioutil.ReadAll(r)
```
mcarmonaa: Is it ok to load all the content in memory for all the blobs?

ajnavarro: Only one file at a time is ok I think.

erizocosmico: If the connection sends the rows in streaming, it should only send 1 at a time, which is ok. If not, it can be huge, yes.

How should we specify the threshold? It should be configurable.

Why limit to non-binary files only? Maybe you want to get a binary file like an image or something. Even if we're gonna use it for code analysis, this is general purpose and someone might want to idk, query for images and get their contents.

ajnavarro: we can leave for the first version two constants, one to only get binary files, and another with a threshold to do not send files bigger than X MB.

erizocosmico: Should we at least have some environment variables for now with a default?

ajnavarro: Ok, but just for now.

ajnavarro: Open an issue for configuration: https://github.com/src-d/go-mysql-server/issues/109

erizocosmico:
Fixed @ajnavarro @mcarmonaa

Issue #174, opened 2018-03-14
Queries can not be cancelled


jfontan:
After cancelling a query it continues running in the server.

jfontan:
When a query is cancelled its callback is not immediately called. The same happens when the connection is lost and `Handler.ConnectionClosed` is not called until some timeout passes.

To cancel a query the client sends a `KILL` command:

```sql
KILL QUERY 1
```
Queries should be registered by query ID and cancelled when this message is received. This has to be done in `go-mysql-server`.

erizocosmico:
Yup, this should be done in the `SessionManager` inside `go-mysql-server`.

I'm assuming the IDs are just an auto-incrementable number with the number of queries executed so far, right?

jfontan:
ID is ConnectionID. I have already working code to save connection structures to close the connection when needed (`KILL CONNECTION`).

erizocosmico:
I mean in `KILL QUERY FOO`

jfontan:
There are two flavours of `KILL`. Both kill the query but one of them (`KILL CONNECTION`) also kills the connection.

https://dev.mysql.com/doc/refman/5.7/en/kill.html

PR #177, opened 2018-03-14
Use session and support query cancelation

jfontan:
`KILL QUERY` command implemented here: https://github.com/src-d/go-mysql-server/pull/131 

Fixes: #174


blobs.go
```diff
@@ -65,10 +63,10 @@ func (r *blobsTable) TransformExpressionsUp(f func(sql.Expression) (sql.Expressi
 	return r, nil
 }
 
-func (r blobsTable) RowIter(_ sql.Session) (sql.RowIter, error) {
+func (r blobsTable) RowIter(session sql.Session) (sql.RowIter, error) {
 	iter := new(blobIter)
 
-	repoIter, err := NewRowRepoIter(r.pool, iter)
+	repoIter, err := NewRowRepoIter(session, iter)
```
erizocosmico: since the only thing the row repo iter needs from the session is the pool, what do you think about keeping it how it was and just extracting the pool from the session in RowIter? It's a bit of repeated code over all RowIter's, but I feel a bit odd tht RowRepoIter needs the session

jfontan: It also needs the context to cancel the iterator:

https://github.com/jfontan/gitquery/blob/626a886bcfb598eaca5195a02074a2fc9a550d12/repository_pool.go#L233

erizocosmico: Ah, ok. Then ignore my comment 😄

jfontan:
I've just added lock to `RowRepoIter.setError` that was unprotected.

jfontan:
Rebased, updated `go-mysql-server` and changed code to be compatible with the new version. Can you take a look @erizocosmico, @ajnavarro, @mcarmonaa?

internal/function/is_remote_test.go
```diff
@@ -28,7 +28,11 @@ func TestIsRemote(t *testing.T) {
 	for _, tt := range testCases {
 		t.Run(tt.name, func(t *testing.T) {
 			require := require.New(t)
-			val, err := f.Eval(sql.NewBaseSession(context.TODO()), tt.row)
+
+			session := sql.NewBaseSession()
```
erizocosmico: You can use `sql.NewEmptyContext()` when it's just for a test

PR #179, opened 2018-03-15
 it: commits that appear on more than one reference

### LINKED ISSUE #159: Add integration test to check the number of commits that appears on more than X references.

erizocosmico:
Closes #159 

Depends on #178 
Depends on https://github.com/src-d/go-mysql-server/pull/103

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -110,6 +110,25 @@ func TestIntegration(t *testing.T) {
 				{int32(1), int32(3), path, "daniel@lordran.local"},
 			},
 		},
+		{
+			`SELECT * FROM (
+				SELECT COUNT(c.hash) AS num, c.hash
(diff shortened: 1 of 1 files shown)


integration_test.go
```diff
@@ -71,6 +105,7 @@ func TestIntegration(t *testing.T) {
 			_, iter, err := engine.Query(session, tt.query)
 			require.NoError(err)
 			rows, err := sql.RowIterToRows(iter)
+			spew.Dump(rows)
```
jfontan: This gives a lot of output. If used it should only print the output on error.

erizocosmico: Shit, that was a leftover 😞

erizocosmico:
Fixed @jfontan

PR #182, opened 2018-03-16
Store a list of paths instead of instantiated reposititories

### LINKED ISSUE #180: Store a list of repository paths instead of opened go-git repositories
go-git is not thread safe and the current method can lead to use the same repository more than once at the same time. Change RepositoryPool to hold repository paths and create go-git repos in the iterator.

jfontan:
Fixes #180

### DIFF
--- a/remotes_test.go
+++ b/remotes_test.go
@@ -45,7 +45,7 @@ func TestRemotesTable_RowIter(t *testing.T) {
 
 	pool := remotes.pool
 	repository, err := pool.GetPos(0)
-	require.Nil(err)
+	require.NoError(err)
 
 	repo := repository.Repo
 
@@ -59,16 +59,16 @@ func TestRemotesTable_RowIter(t *testing.T) {
 	}
 
 	_, err = repo.CreateRemote(&config)
-	require.Nil(err)
+	require.NoError(err)
 
 	rows, err := sql.NodeToRows(sql.NewBaseSession(context.TODO()), table)
-	require.Nil(err)
+	require.NoError(err)
 	require.Len(rows, 3)
 
 	schema := table.Schema()
 	for idx, row := range rows {
 		err := schema.CheckRow(row)
-		require.Nil(err, "row %d doesn't conform to schema", idx)
+		require.NoError(err, "row %d doesn't conform to schema", idx)
 
 		if row[1] == "my_remote" {
 			urlstring, ok := row[2].(string)
--- a/remotes_test.go
+++ b/remotes_test.go
@@ -16,7 +16,7 @@ import (
 func TestRemotesTable_Name(t *testing.T) {
 	require := require.New(t)
 
-	f := fixtures.Basic().One()
+	f := fixtures.ByTag("worktree").One()
 	table := getTable(require, f, remotesTableName)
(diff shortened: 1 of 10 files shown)


repository_pool.go
```diff
@@ -25,34 +25,53 @@ func NewRepository(id string, repo *git.Repository) Repository {
...
-	repositories []Repository
+	repositories map[string]string
+	idOrder      []string
 }
 
 // NewRepositoryPool initializes a new RepositoryPool
 func NewRepositoryPool() RepositoryPool {
-	return RepositoryPool{}
+	return RepositoryPool{
+		repositories: make(map[string]string),
```
mcarmonaa: Is it expected for the moment that a repo has only one path?

repository_pool.go
```diff
@@ -25,34 +25,53 @@ func NewRepository(id string, repo *git.Repository) Repository {
...
+	}
+
+	return NewRepository(id, repo), nil
+}
+
+// RepositoryPool holds a pool git repository paths and
 // functionality to open and iterate them.
 type RepositoryPool struct {
-	repositories []Repository
+	repositories map[string]string
```
erizocosmico: what is the key and the value to this? I saw in the tests the key is the repo path, but what's the value?

repository_pool.go
jfontan: Yes, in the filesystem the repo should be only once.

repository_pool.go
jfontan: The key is the ID. The idea is not to use the full path as an ID. For example if we have the path `repos/00/001122334455` the ID could be `001122334455`. This is the same ID that appears as `repository_id` column.

remotes_test.go
```diff
@@ -29,23 +29,23 @@ func TestRemotesTable_Name(t *testing.T) {
...
 	table := getTable(require, f, remotesTableName)
 
 	remotes, ok := table.(*remotesTable)
 	require.True(ok)
 
 	pool := remotes.pool
-	repository, ok := pool.GetPos(0)
-	require.True(ok)
+	repository, err := pool.GetPos(0)
+	require.Nil(err)
```
ajnavarro: NoError

jfontan: Changed

repository_pool.go
erizocosmico: What if we have `/repos/foo` and `/repos/bar/foo`? Isn't that a bit misleading? Not everyone will have the repositories bucketed, so name clashes may happen.

jfontan: We can either change `AddGit` to add another number of directories or set the `ID` manually:

```
pool.Add("some_fancy_identifier", "/storage/bricks/83/repos/846283")
```

Truncating the name to create the ID is just quick fix for now to have a `repository_id` that is not the full path. In the future that identifier should be unique and really identify the repo. For example with a UUID. If we use the full path and the repository is moved the the ID will change and this may lead to problems.

I can set the full path for now if you think this will be less error prone.

jfontan:
Rebased and fixed an integration test. It was using full path as ID.

repository_pool.go
jfontan: Now ID is the full path.

PR #183, opened 2018-03-16
history_idx: do not visit a parent more than once

jfontan:
There are cases where it can enter a cycle

### DIFF
--- a/internal/function/history_idx.go
+++ b/internal/function/history_idx.go
@@ -125,6 +125,7 @@ func (f *HistoryIdx) repoHistoryIdx(repo *git.Repository, start, target plumbing
 	// history. Because the frame keeps track of which was its index, we can
 	// return accurate indexes even if there are multiple branches.
 	stack := []*stackFrame{{0, 0, []plumbing.Hash{start}}}
+	visitedHashes := make(map[plumbing.Hash]struct{})
 
 	for {
 		if len(stack) == 0 {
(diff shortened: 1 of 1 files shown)


internal/function/history_idx.go
```diff
@@ -153,7 +160,17 @@ func (f *HistoryIdx) repoHistoryIdx(repo *git.Repository, start, target plumbing
 		}
 
 		if c.NumParents() > 0 {
-			stack = append(stack, &stackFrame{frame.idx + 1, 0, c.ParentHashes})
+			newParents := make([]plumbing.Hash, 0, c.NumParents())
+			for _, h = range c.ParentHashes {
+				_, ok = visitedHashes[h]
+				if !ok {
```
erizocosmico: you can inline that as `if _, ok := visitedHashes[h]; !ok`

internal/function/history_idx.go
```diff
@@ -133,7 +134,13 @@ func (f *HistoryIdx) repoHistoryIdx(repo *git.Repository, start, target plumbing
 
 		frame := stack[len(stack)-1]
 
-		c, err := repo.CommitObject(frame.hashes[frame.pos])
+		h := frame.hashes[frame.pos]
+		_, ok := visitedHashes[h]
+		if !ok {
```
erizocosmico: you can inline that as `if _, ok := visitedHashes[h]; !ok`

ajnavarro:
The go-mysql-server broke API comp. We need another PR implementing String method and rebase here.

erizocosmico:
Travis was failing because you were caught in the middle of making it compatible with the new go-mysql-server. Restarted the build.

PR #194, opened 2018-03-22
 internal/function: implement language function 

### LINKED ISSUE #192: implement UDF that gets the language from the content of the blob and/or its path

erizocosmico:
Closes #192 

Can be used as:

```sql
language(path)

OR

language(path, blob)
```


erizocosmico:
Rebased

internal/function/language.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return fn(&Language{left, right})
+}
+
+// Eval implements the Expression interface.
+func (f *Language) Eval(session sql.Session, row sql.Row) (interface{}, error) {
```
jfontan: Eval now uses `sql.Context` instead of `Session`

erizocosmico: Oh, right, the update is already on master

erizocosmico:
Rebased

PR #195, opened 2018-03-23
 internal/function: implement uast and uast_xpath functions

### LINKED ISSUE #193: Implements UDF that returns the UAST from a file
UDF that internally uses bblfsh. It should be like:

- `uast(content:BLOB, [lang:TEXT],[xpath:TEXT]) BLOB`

Also we will need another UDF to query again an uast

- `uast_xpath(uast:BLOB,xpath:TEXT) BLOB`

erizocosmico:
Closes #193


erizocosmico:
Tests fixed, required some work done on the CI for bblfsh/client-go to work.

session.go
```diff
@@ -11,20 +16,89 @@ import (
...
+	for {
+		if attempts > bblfshMaxAttempts {
+			return nil, ErrBblfshConnection.New()
+		}
+
+		switch s.bblfshClient.GetState() {
+		case connectivity.Ready, connectivity.Idle:
+			return s.bblfshClient, nil
+		case connectivity.Connecting:
+			time.Sleep(100 * time.Millisecond)
```
jfontan: Shouldn't we reset or decrease attempt counter? If it's connection at attempt 10 it will exit the next loop.

erizocosmico: That could lead to it being in an infinite loop if it's connecting, then fails, then connecting, then fails. We could set a maximum total retries and reset only a local counter, tho

erizocosmico:
Done @jfontan

erizocosmico:
Rebased

erizocosmico:
Rebased and CI fixed.

erizocosmico:
@ajnavarro integration test added

PR #196, opened 2018-03-26
perf: add caches to commit_has_blob and commit_has_tree

### LINKED ISSUE #191: Optimize commit_has_blob and commit_has_tree
Find a way to optimize this Function, using cache or another kind of index.

erizocosmico:
Closes #191 

### Commit has tree

Before:

```
BenchmarkCommitHasTree/commit_has_tree-4                    3000            519234 ns/op          297322 B/op       1276 allocs/op
PASS
ok      github.com/src-d/gitquery/internal/function     2.154s
```

After:

```
BenchmarkCommitHasTree/commit_has_tree-4                 1000000              1036 ns/op             240 B/op          7 allocs/op
PASS
ok      github.com/src-d/gitquery/internal/function     1.593s
```

### Commit has blob

Before:

```
BenchmarkCommitHasBlob/commit_has_blob-4                     500           2605190 ns/op         1143268 B/op       1768 allocs/op
PASS
ok      github.com/src-d/gitquery/internal/function     2.105s
```

After:

```
BenchmarkCommitHasBlob/commit_has_blob-4                 1000000              1013 ns/op             240 B/op          7 allocs/op
PASS
ok      github.com/src-d/gitquery/internal/function     1.559s
```

**DISCLAIMER:** benchmark here is biased because the same commits keep being asked again and again. In a regular production query the improvement would likely be much lower than it is here. Still an improvement, nonetheless.

### DIFF
--- a/internal/function/commit_has_blob.go
+++ b/internal/function/commit_has_blob.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/hashicorp/golang-lru"
+
 	"github.com/src-d/gitquery"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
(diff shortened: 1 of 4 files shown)


erizocosmico:
To no add it twice, the commit adding golang-lru to vendor is in #198

internal/function/commit_has_blob.go
```diff
@@ -14,13 +16,18 @@ import (
...
 	commitHash sql.Expression
 	blob       sql.Expression
+	cache      *lru.ARCCache
 }
 
+const commitHasBlobCacheSize = 200
+
 // NewCommitHasBlob creates a new commit_has_blob function.
 func NewCommitHasBlob(commitHash, blob sql.Expression) sql.Expression {
+	cache, _ := lru.NewARC(commitHasBlobCacheSize)
```
mcarmonaa: Maybe it's a silly question but, isn't there any problem using `ARC cache` given that IBM has the patent of it?

erizocosmico: Actually, not silly at all. I think we should move it to `TwoQueueCache`, then. What do you think? @ajnavarro @jfontan

erizocosmico:
Changed from ARCCache to TwoQueueCache

PR #198, opened 2018-03-26
internal/function: add cache for history_idx

### LINKED ISSUE #190: Optimize history_idx UDF
Find a way to optimize this Function, using cache or another kind of index.

erizocosmico:
Fixes #190 

This commit adds a LRU cache to the history_idx function, caching the indexes in which a commit is from another in the history.

Before:

```
BenchmarkHistoryIdx/history_idx-4                   2000            789749 ns/op          415896 B/op       1519 allocs/op
PASS
ok      github.com/src-d/gitquery/internal/function     2.189s
```

After:

```
BenchmarkHistoryIdx/history_idx-4                1000000              1063 ns/op             256 B/op          9 allocs/op
PASS
ok      github.com/src-d/gitquery/internal/function     1.607s
```

**DISCLAIMER:** benchmark here is biased because the same commits keep being asked again and again. In a regular production query the improvement would likely be much lower than it is here. Still an improvement, nonetheless.

### DIFF
--- a/internal/function/history_idx.go
+++ b/internal/function/history_idx.go
@@ -17,14 +17,14 @@ import (
 // of another commit.
 type HistoryIdx struct {
 	expression.BinaryExpression
-	cache *lru.ARCCache
+	cache *lru.TwoQueueCache
 }
 
 const historyIdxCacheSize = 300
 
 // NewHistoryIdx creates a new HistoryIdx udf.
 func NewHistoryIdx(start, target sql.Expression) sql.Expression {
-	cache, _ := lru.NewARC(historyIdxCacheSize)
+	cache, _ := lru.New2Q(historyIdxCacheSize)
 	return &HistoryIdx{expression.BinaryExpression{Left: start, Right: target}, cache}
 }
 
--- a/internal/function/history_idx.go
+++ b/internal/function/history_idx.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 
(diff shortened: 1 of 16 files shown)


internal/function/history_idx.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/hashicorp/golang-lru"
```
smola: I guess this needs updating `Gopkg.*`.

erizocosmico: Good catch!

erizocosmico:
@smola golang-lru added to vendor

internal/function/history_idx.go
```diff
@@ -15,17 +17,25 @@ import (
...
 	expression.BinaryExpression
+	cache *lru.ARCCache
 }
 
+const historyIdxCacheSize = 300
+
 // NewHistoryIdx creates a new HistoryIdx udf.
 func NewHistoryIdx(start, target sql.Expression) sql.Expression {
-	return &HistoryIdx{expression.BinaryExpression{Left: start, Right: target}}
+	cache, _ := lru.NewARC(historyIdxCacheSize)
```
mcarmonaa: Same question than in #196 about the IBM's patent on the `ARC`

erizocosmico:
Changed from ARCCache to TwoQueueCache

erizocosmico:
Rebased

PR #201, opened 2018-04-06
 *: implement rule to squash chainable tables into a single one 

### LINKED ISSUE #197: Performance: Reduce joins to one node to improve performance
Create a rule to transform the LogicalPlan, getting joins between git tables and transforming them to only one node. Doing this we only need to iterate once per repository per join, instead of doing a join between all the repositories (like we are doing on engine project).

As in engine, we should only support Projects, Filter and Table nodes below Joins. If another node type is there, the Join node cannot be optimized.

erizocosmico:
Fixes #197

This commit introduces the new rule to squash inner joins between tables that are chainable. This rule is yet experimental and thus can be used by using the `UNSTABLE_SQUASH_ENABLE` environment variable until it is proven to be stable and the results match the ones retrieved without this optimisation.

What is a join between chainable tables? That's an inner join between two tables that can be joined using some logic in which one table records are retrieved from some data in the other table rows because there is some sort of relationship between them. To assert such logic can be applied, the join needs to have certain filters in its condition.

- `repositories` is chainable with `remotes` iff `repositories.id = remotes.repository_id`
- `repositories` is chainable with `refs` iff `repositories.id = refs.repository_id`
- `remotes` is chainable with `refs` iff `refs.repository_id = remotes.repository_id`
- `refs` is chainable with `commits` iff `refs.hash = commits.hash`, `history_idx(refs.hash, commits.hash) >= 0` or `0 <= history_idx(refs.hash, commits.hash)`
- `refs` is chainable with `tree_entries` iff `commit_has_tree(refs.hash, tree_entries.tree_hash)`
- `refs` is chainable with `blobs` iff `commit_has_blob(refs.hash, blob.hash)`
- `commits` is chainable with `tree_entries` iff `commit_has_tree(commits.hash, tree_entries.tree_hash)` or `commits.tree_hash = tree_entries.tree_hash`
- `commits` is chainable with `blobs` iff `commit_has_blob(commits.hash, blob.hash)`
- `tree_entries` is chainable with `blobs` iff `tree_entries.entry_hash = blobs.hash`

Since chains are built hierarchically using the table hierarchy (repositories -> remotes -> refs -> commits -> tree_entries -> blobs), any chain can be joined with any table iff the last part of the chain and that table match the conditions stated before.

To allow this, there are several iterators, some for the same table, depending on the position in the chain that table is, that are composed in the analyzer rule.

Even if the chain is built hierarchically and thus it has a different schema ordering than the original schema, a mapping between the chain schema and the original one is computed so that the result of the chain is exactly the same as the one without the optimisation.

Notes about the experimental status of this feature:

- Lacks some integration tests that will be done on subsequent PRs.
- Tree entries table is currently under examination and depending on what we decide, the results it returns right now might need to change.
- Probably better to test it thoroughly and do some performance measurements before making this the default behaviour.

Now, take some coffee, get comfy in your chair and be patient, 'cause this is the mother of all PRs.


internal/rule/squashjoins.go
```diff
@@ -0,0 +1,948 @@
...
+	tables []sql.Table,
+	filters, columns []sql.Expression,
+) (sql.Node, error) {
+	tableNames := orderedTableNames(tables)
+
+	var iter gitquery.ChainableIter
+	var err error
+	for _, t := range tableNames {
+		switch t {
+		case "repositories":
```
ajnavarro: Maybe you should use the already existing variable for table names.

internal/rule/squashjoins.go
```diff
@@ -0,0 +1,948 @@
...
+					return nil, err
+				}
+				iter = gitquery.NewAllReposIter(f)
+			default:
+				return nil, errInvalidIteratorChain.New("repositories", iter)
+			}
+		case "remotes":
+			switch it := iter.(type) {
+			case gitquery.ReposIter:
+				var f sql.Expression
```
ajnavarro: I don't know if it's worth it or just for code clarity, but this logic is almost duplicated on `case "refs"`

ajnavarro: Just the same comment for all the duplicated cases.

internal/rule/squashjoins.go
```diff
@@ -0,0 +1,948 @@
...
+		num, ok := result.(int64)
+		if !ok {
+			return false
+		}
+
+		return num == n
+	}
+}
+
+var errColumnNotFound = errors.NewKind("column %s.%s not found in schema")
```
ajnavarro: we should use ErrColumnNotFound from `go-mysql-server`: https://sourcegraph.com/github.com/src-d/go-mysql-server/-/blob/sql/analyzer/rules.go#L32

internal/rule/squashjoins.go
erizocosmico: Yep, but there's no way to do that generically without repeating some code because the types and the params of everything inside those cases are not the same.

erizocosmico:
Fixed @ajnavarro

erizocosmico:
**UPDATE:**

There was an inconsistency between the results returned by the iterators and the ones using the tables and joins because the tables look for all trees and visit them. Instead, the iterator chained with commits only visited the main tree. Now, there is another iterator for this case that visits the main tree and puts other trees that are found in a queue to visit them as well. This mimics the behaviour of joinin commits and tree entries.

ajnavarro:
We should document `UNSTABLE_SQUASH_ENABLE` on README.

erizocosmico:
Rebased after rename.

PR #204, opened 2018-04-10
*: rename gitquery to gitbase

erizocosmico:
After this, we can rename the repo.


bzz:
screen shot 2018-04-10 at 12 41 39 pm

PR #206, opened 2018-04-10
add dockerfile

### LINKED ISSUE #203: Release: add Docker image
To simplify a new user's experience, it would be useful to have a Docker image with gitbase pushed somewhere, as one of the release artefacts of the project.

erizocosmico:
Closes #203 

Publishing with CI will be done on another PR.

Final image is not alpine because we're using cgo and I could not get it to work in any way. If someone knows how to run CGO binaries on alpine, please come forward.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,16 @@ ENV GITBASE_USER=gitbase
 ENV GITBASE_PASSWORD=""
 EXPOSE 3306
 
-RUN apt-get update && apt-get install libxml2-dev git -y \
+ENV TINI_VERSION v0.17.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini
+
+RUN apt-get update \
+    && apt-get install libxml2-dev git -y \
     && apt-get clean \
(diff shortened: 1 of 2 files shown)


jfontan:
We can revisit the base image of it in the future if needed but it's probably ok as is for now.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN cd vendor/gopkg.in/bblfsh/client-go.v2 && make dependencies
+RUN go install -v github.com/src-d/gitbase/...
+
+FROM ubuntu:16.04
+
+COPY --from=builder /go/bin/gitbase /bin
+RUN mkdir -p /opt/repos
+
+ENV GITBASE_USER=gitbase
+ENV GITBASE_PASSWORD=""
```
kuba--: just because I'm curious - do we rely on some image where `GITBASE_PASSWORD` is exported or we're just using empty password?

erizocosmico: We're using no password by default and then you can add your own password when you run it using `-e GITBASE_PASSWORD="your password"`

erizocosmico:
@jfontan yep, probably in the future we could run it as non-root user, as non pid 1 and so on

erizocosmico:
Added tini to run gitbase as non pid 1

PR #212, opened 2018-04-11
*: tests for squash vs no squash correctness

erizocosmico:
This commit implements a test suite to ensure results using the squash optimization are the exact same as the ones not using it.
It also fixes two of these differences:
- Without squashing, remote iterator did not return a string but another type for the refspecs.
- Joining blobs with refs or commits could lead to extra rows because blobs could appear more than once per commit or ref. Now they only appear once.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ _testmain.go
 *.exe
 *.test
 *.prof
+gitbase
 
 # CI
 .ci/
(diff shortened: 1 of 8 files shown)


ajnavarro:
Maybe we should change the rest of the tests to use always the squash rule? It's something that in the future will be always active.

erizocosmico:
All the integration tests already use squash (and no squash)

PR #219, opened 2018-04-12
*: don't read blob content if it's not used

### LINKED ISSUE #209: Excesive memory consumption
Executing the following query with 3 small repositories (go-billy, go-git, gitbase) using the ScuashJoins rule:

```
SELECT COUNT(c.hash), c.hash
FROM refs r
INNER JOIN commits c
	ON r.name = 'HEAD' AND history_idx(r.hash, c.hash) >= 0
INNER JOIN blobs b
	ON commit_has_blob(c.hash, b.hash)
GROUP BY c.hash
```

The amount of memory consumed is huge (16GB + part of the Swap). Even when the query finish, the memory is still occupied.

erizocosmico:
Fixes #209 

Blob contents should not be read unless the blobs.content column is specifically used in the query (either in direct use or in a projection) to save memory. Operations such as aggregations may use a huge amount of memory without this optimisation.

Thanks to the column pushdown we are able to know if the column is in use, so if it is, just flag the iterators to let them know they should read the content.


blobs.go
```diff
@@ -238,3 +244,22 @@ func isBinary(blob *object.Blob) (bool, error) {
 		}
 	}
 }
+
+func shouldReadContent(columns []sql.Expression) bool {
```
ajnavarro: This maybe should be a rule that adds a boolean to blobs table?

erizocosmico: I think the rule will end up being more expensive than just walking a few expressions. This is done once when the iterator is created, the rule will probably trigger a new pass of the analyzer (since the analyzer creates new nodes and the tree will change) and all its rules, which may end up being way more expensive than checking this here.

Note that this part is only used for non-squash. In squash, it's done in the rule because iterators are built on analysis time.

PR #220, opened 2018-04-12
gitbase: skip hash references that do not point to commits

### LINKED ISSUE #213: Squashed Join: Object not found error
`Object not found error` happens sometimes using the squash joins rule. It appears to be because references can point to any object type, not only commits. Check if we are ignoring `object not found` error when a reference is pointing to a not-a-commit object.

erizocosmico:
Fixes #213

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	errors "gopkg.in/src-d/go-errors.v1"
 	git "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/filemode"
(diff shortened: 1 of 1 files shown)


iterator.go
```diff
@@ -791,6 +791,10 @@ func (i *refCommitsIter) Advance() error {
 				return io.EOF
 			}
 
+			if err == plumbing.ErrObjectNotFound {
```
kuba--: lets reverse order of all these checks (why do we check EOF and Object not found if an error is nil?
```go
if err != nil {
if err == io.EOF {
 				i.refs = nil
  				return io.EOF
  			}
if err == plumbing.ErrObjectNotFound {
  continue
}
return err
```

ajnavarro:
I think we should do something like here: https://github.com/src-d/borges/blob/master/git.go#L219

Object not found can be for other reasons too.

erizocosmico:
@ajnavarro Sounds good. Should we resolve the tag to a commit like this does, too, instead of skipping it?

erizocosmico:
Fixed (waiting on Antonio to tell if we should resolve the tags to a commit before adding that)

ajnavarro:
@erizocosmico yes, we should resolve tags too to commits if possible

Issue #215, opened 2018-04-12
panic on RowRepoIter


smola:
Query:

```sql
SELECT c.message
FROM
    commits c
    JOIN refs r ON r.hash = c.hash
WHERE
    r.name = 'refs/heads/HEAD';
```

Panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xafdf73]

goroutine 39 [running]:
github.com/src-d/gitbase.(*rowRepoIter).rowReader(0xc420216850, 0x1)
	/go/src/github.com/src-d/gitbase/repository_pool.go:293 +0x193
created by github.com/src-d/gitbase.NewRowRepoIter
	/go/src/github.com/src-d/gitbase/repository_pool.go:214 +0x277
opening "/opt/repos" repository...
time="2018-04-12T07:10:36Z" level=info msg="NewConnection: client 1"
time="2018-04-12T07:17:00Z" level=info msg="NewConnection: client 2"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xafdf73]

goroutine 69 [running]:
github.com/src-d/gitbase.(*rowRepoIter).rowReader(0xc4201fa150, 0x1)
	/go/src/github.com/src-d/gitbase/repository_pool.go:293 +0x193
created by github.com/src-d/gitbase.NewRowRepoIter
	/go/src/github.com/src-d/gitbase/repository_pool.go:214 +0x277
```

smola:
Here's the problem: https://github.com/src-d/gitbase/blob/847f97f942875e8979a349032d1c8f8d701080e4/repository_pool.go#L279

ajnavarro:
@smola Are you using `UNSTABLE_SQUASH_ENABLE=1`?

smola:
@ajnavarro No.

PR #221, opened 2018-04-12
Catch errors creating new iteratos in rowReaders

jfontan:
Fixes: #215

### DIFF
--- a/repository_pool.go
+++ b/repository_pool.go
@@ -276,7 +276,18 @@ func (i *rowRepoIter) rowReader(num int) {
 	defer i.wg.Done()
 
 	for repo := range i.repos {
-		iter, _ := i.iter.NewIterator(repo)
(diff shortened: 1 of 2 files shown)


repository_pool.go
```diff
@@ -276,7 +276,17 @@ func (i *rowRepoIter) rowReader(num int) {
...
-		iter, _ := i.iter.NewIterator(repo)
+		iter, err := i.iter.NewIterator(repo)
+		if err != nil {
+			// guard from possible previous error
+			select {
+			case <-i.done:
+				return
+			default:
+				i.setError(err)
+				close(i.done)
```
smola: You need to `continue` to the next loop iteration. Otherwise, we'll still get a panic because of the `iter == nil` and never get the error.

jfontan:
Added test for this case and fixed the problem pointed by @smola.

PR #223, opened 2018-04-12
*: add logging

### LINKED ISSUE #217: Add logs
Add logs to gitbase, to be able to see what is happening on execution, and improve error reporting.

erizocosmico:
Closes #217

### DIFF
--- a/cmd/gitbase/server.go
+++ b/cmd/gitbase/server.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/src-d/gitbase"
 	"github.com/src-d/gitbase/internal/function"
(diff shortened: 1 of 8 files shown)


erizocosmico:
Sure @ajnavarro

cmd/gitbase/server.go
```diff
@@ -85,5 +95,7 @@ func (c *CmdServer) Execute(args []string) error {
 		return err
 	}
 
+	logrus.Debug("starting server")
```
ajnavarro: maybe we should add a defer logrus.Debug("server started")

erizocosmico: Start never returns until the connection stops, so a defer would actually be "stopping server" 😛

ajnavarro: oh you're right

PR #224, opened 2018-04-12
 gitbase: add support for siva files 

erizocosmico:
Depends on #223 (will probably rebase and remove the dependency tomorrow)

Add support in RepositoryPool for handling siva files as repositories.

Added a new `AddSivaDir` method to `RepositoryPool`, which adds all the siva files inside that directory (a siva file is considered to be any file whose extension is `.siva`) and the direct child directories to allow bucketing.

Also some minor changes related to the RepositoryPool:
- NewRepositoryFromPath returns now *Repository for convenience.
- NewRepositoryPool returns now *RepositoryPool, as it always needs to be used with a pointer.
- Fixed golint warning about returning an unexported type in an exported method, which may be annoying by replacing the type with `sql.RowIter`.


cmd/gitbase/server.go
```diff
@@ -44,12 +45,12 @@ func (c *cmdServer) buildDatabase() error {
 		logrus.WithField("dir", c.Git).Debug("added folder containing git repositories")
 	}
 
-	var err error
+	c.pool = gitbase.NewRepositoryPool()
```
ajnavarro: so, now you can have mixed repositories on the same execution, standard ones and siva ones?

erizocosmico: Yup. I can restrict that if we want to only have one type

ajnavarro: it's ok, LGTM!

PR #229, opened 2018-04-13
benchmark: implement gitbase benchmarks

### LINKED ISSUE #207: Benchmark gitbase
We need a reproducible way to run a set of queries on a specific subset of repositories, to be able to compare gitbase versions (performance, CPU, memory and so on).

The set of queries is not complete but here is a draft: https://gist.github.com/smola/7013c2ce00788a36d9b57c66521a340d

For the data, we can start using a 10% of the PGA for testing. This 10% should contain always the same siva files.

erizocosmico:
Closes #207 

I'm putting it here but don't know if we should keep this in another repository or what. Dataset takes a few hours to download (at least with my connection) so we might wanna look into putting it into the repo somehow (@kuba-- suggested using git lfs).

It's also not running it on CI for obvious reasons, so until we decide where it's going to run it can stay like this.

Squash rule is always active for this because the queries would never end without it.

**Known issues:**

- [ ] `object not found` (yes, the error is _that_ descriptive) on the last benchmark.
- [ ] siva files are rooted repositories so references are something like `UUID/refname`, we might want to split them into proper repositories or it's gonna be hard to query (hence the 0 rows in everything that filters by `refs/heads/HEAD`.
- [ ] Adapted some queries from @smola's gist to make them work on engine and removed the ones that can not be done. Top tokens per repo, for example, we have no way to join blobs and repos (though we could build someting like `repository_has_blob` to do it).


benchmark/benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (b benchmark) run(pool *gitbase.RepositoryPool, e *sqle.Engine) error {
+	info("[RUN] %s", b.name)
+	info(b.query)
+	session := gitbase.NewSession(pool)
+	ctx := sql.NewContext(context.TODO(), sql.WithSession(session))
+	start := time.Now()
+
+	_, iter, err := e.Query(ctx, string(b.query))
```
mcarmonaa: Why is this casting necessary? Isn't `b.query` already a string?

erizocosmico: Leftover, before the benchmark type was an aliased string

jfontan:
PGA server is located in USA and that makes the download slower. Maybe using git lfs gives us a closer and faster server (CDN?). It's worth checking.

erizocosmico:
Build should be fixed now

erizocosmico:
Should we merge this or keep it until we know where it's going to run and how to fetch the dataset and so on?

kuba--:
I like git-lfs. I used it for a while to keep binary dependencies. You can fetch them like regular files and it gives you versioning.

smola:
@erizocosmico  Update on this: we'll be running this on Jenkins. We will be able to access a PGA copy from there at a given path. So if this has the option of just using a predefined path, we'll be able to automate it.

erizocosmico:
@smola Right now is on `$PWD/.pga/siva/latest`. But could be changed to something else if needed

PR #230, opened 2018-04-13
Add env var to skip go-git errors

### LINKED ISSUE #214: Add environment variable to allow git errors and keep processing
Add configuration to be able to ignore errors from go-git and continue iterating the query result.

jfontan:
Fixes #214

### DIFF
--- a/internal/function/history_idx.go
+++ b/internal/function/history_idx.go
@@ -5,10 +5,11 @@ import (
 	"io"
 
 	"github.com/hashicorp/golang-lru"
+	"github.com/sirupsen/logrus"
 
 	"github.com/src-d/gitbase"
-	git "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 )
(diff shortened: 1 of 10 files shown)


erizocosmico:
`iterator.go` file has a bunch more of possible places where there could be go-git errors

jfontan:
@erizocosmico yes, I'm on it right now

jfontan:
There are no new tests for functions as I could not find a way to make git repository return errors. Deleting files from the path makes it return `ErrObjectNotFound` and it's skipped.

There are also no new tests in `iterator_test.go` `TestRecursiveTreeFileIter` for the same reason.

smola:
@jfontan Maybe you can remove the repository in the middle of an iteration.

jfontan:
I did that, removing index and packfile but the error (`ErrObjectNotFound`) is already skipped by functions.

internal/function/history_idx.go
```diff
@@ -133,6 +184,10 @@ func (f *HistoryIdx) repoHistoryIdx(repo *git.Repository, start, target plumbing
 	}
 
 	if err != nil {
+		log.WithFields(logrus.Fields{
+			"error":  err,
+			"commit": target.String(),
+		}).Error("could not get commit")
```
erizocosmico: shouldn't we skip this too if `SkipGitErrors` is `true`?

iterator.go
```diff
@@ -715,11 +771,23 @@ func (i *commitsIter) Close() error {
...
 
+	commits, err := repo.Repo.CommitObjects()
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"repo":  repo.ID,
+			"error": err,
+		}).Error("unable to get commit iterator")
+
+		if session.SkipGitErrors {
+			return nil, err
```
erizocosmico: shouldn't we skip this error?

PR #236, opened 2018-04-17
Add rules to the Makefile for bblfsh commands

mcarmonaa:
Since gitbase now requires a bblfsh server, I though it would be good have the bblfsh commands as rules in the makefile to simplify the process.


mcarmonaa:
It's is already defined in `.travis.yml`, the goal it's just for developing use.

mcuadros:
If is already in the travis, removed from the Makefile.

mcarmonaa:
Closing it, the Makefile is not a good place for that. Should be documented

PR #243, opened 2018-04-18
cli: small refactor to align the other products

mcuadros:
This is a small refactor to align it with other company product like:
https://github.com/bblfsh/bblfshd/blob/master/cli/bblfshctl/

Also will be nice add some tests to the commands.


smola:
@mcuadros  `cmd` is way more common than `cli` in the Go community. Shouldn't we change it the other way around? `cli -> cmd`.

There was some previous discussion here: https://github.com/src-d/guide/pull/196#discussion_r181054490

mcuadros:
Its already, in many places, and don't worth to changed, and `cli` is a more meaningful name that `cmd`.

But this changes are more related to the scaffolding than the naming.

jfontan:
If we are not going to have more sub-commands I would dump them. By default it starts a server and `--version` prints the version number.

PR #257, opened 2018-05-03
Fix wrong commands path on Makefile

ajnavarro:


bzz:
Tried locally, `make packages` produces the executable binaries now 🎉 

Seems a bit confusing due to https://github.com/src-d/guide/blob/master/engineering/conventions-go.md#executables

erizocosmico:
Shouldn't we instead change the folder name to `cmd` as the engineering guide says?

cmd/gitbase/main.go
```diff
@@ -3,7 +3,7 @@ package main
 import (
 	"os"
 
-	"github.com/src-d/gitbase/cli/gitbase/cmd"
+	"github.com/src-d/gitbase/cmd/gitbase/cmd"
```
erizocosmico: maybe we should move the contents of the cmd folder into the gitbase folder. It does not make much sense to have them in a folder and the path looks weird

ajnavarro:
@erizocosmico done.

cmd/gitbase/server.go
```diff
@@ -1,4 +1,4 @@
-package cmd
+package main
```
smola: Why? Subcommnads are supposed to be implemented in a subpackage.

erizocosmico: Why do they need to be in a subpackage?

mcuadros: Testing.

mcuadros:
Who this PR should the problem of the build?

ajnavarro:
@mcuadros Packfile had a wrong path pointing to the commands folder.

mcuadros:
But this PR is not changing the Makefile

ajnavarro:
That was the first approach, but we decide to do it correctly and change instead the folder to the correct one due to the company guidelines: `cli` to `cmd`.

mcuadros:
But this PR is removing the package for the commands.

ajnavarro:
Moved all commands to `cmd/gitbase/command` package after an offline talk with @mcuadros.

ajnavarro:
Merging to fix the build. If we want to add more changes to the path, we can do it into another PR.

Issue #295, opened 2018-06-04
Use packfile reader when is possible


ajnavarro:
Maybe we can improve objects read performance accessing directly to the packfile.

To be able to do that, we need the offset of the file, that will be provided by the index.

First of all, we need to check the performance impact of that change on a big repository doing a quick benchmark, using directly `Repository` type and packfile `Decoder` type, in the following cases:
  >When random access is needed, use a slice of hashes when using `Repository` and a list of offsets when using `Decoder`

- Read the first blob. 
- Read the first 100 blobs.
- Read 100 blobs, skipping 5 every step.
- Read the last blob.

This benchmark should be done only using go-git, gitbase is not necessary.

If the results using Decoder directly are better than using `Repository`, we can continue and use Decoder directly on some cases.

erizocosmico:
Here's the benchmark: https://github.com/erizocosmico/packfile-read-benchmark

The results of using the decoder is orders of magnitude faster. Note that this is without using cache, I tried reusing both instances of decoder and repository instead of opening one each run and they are slightly similar (except for the last).

erizocosmico:
These are the result if we also open the repository in each iteration of the decoder benchmark:

```
go test -bench=. -benchmem
goos: darwin
goarch: amd64
pkg: github.com/erizocosmico/packfile-read-benchmark
BenchmarkDecoder/first-4         	    2000	   1216390 ns/op	   95273 B/op	     226 allocs/op
BenchmarkDecoder/first_100-4     	     100	  14238049 ns/op	 9215988 B/op	    3537 allocs/op
BenchmarkDecoder/first_100_skip_5-4         	     100	  23771055 ns/op	13205025 B/op	    8680 allocs/op
BenchmarkDecoder/last-4                     	    3000	    871919 ns/op	  286663 B/op	     273 allocs/op
BenchmarkRepository/first-4                 	      10	 145066813 ns/op	60297573 B/op	 1238044 allocs/op
BenchmarkRepository/first_100-4             	      10	 161075898 ns/op	74482648 B/op	 1246176 allocs/op
BenchmarkRepository/first_100_skip_5-4      	      10	 221480822 ns/op	78426575 B/op	 1251132 allocs/op
BenchmarkRepository/last-4                  	      10	 151260601 ns/op	60488844 B/op	 1238090 allocs/op
PASS
ok  	github.com/erizocosmico/packfile-read-benchmark	49.224s
```

As we can see, it's almost the same.

[...]

PR #279, opened 2018-05-16
*: implement indexable interfaces on basic and relation tables

mcarmonaa:
Depends on [src-d/go-mysql-server] PRs:
- https://github.com/src-d/go-mysql-server/pull/181, until this is merged, the `go-mysql-server` dependency can't be updated because of tests failing. Also here is implemented the method `IndexOf` for the schema which is used by the `IndexKeyValueIter` method

- https://github.com/src-d/go-mysql-server/pull/206, until this is merged, the `go-mysql-server` dependency can't be updated because of tests failing.

- https://github.com/src-d/go-mysql-server/pull/193, in this PR `sql.Context` is added to the `sql.Indexable` interfaces. This change is needed to have access to the repository pool from the `sql.Indexable` methods implementation.

@src-d/data-retrieval just have a look to the proposed implementation, if all is ok I'll address the dependency update.

Signed-off-by: Manuel Carmona <manu.carmona90@gmail.com>

PR #298, opened 2018-06-05
gitbase: implement sql.Indexable on all tables

erizocosmico:
Most of the contents in #279 did not survive because they relied on the index being something generic for all tables when in reality, almost every table has its own logic regarding the index storage.

This implementation differs *a lot* from what was initially thought in #295. The way the indexes work made it impossible to work that way.
There is also a lot of code that is inside go-git, but that logic is not exposed. If it's ever exposed we might want to refactor some of that.

To sum it up, the implementation looks like this:

- Every table has its own key value iterator with its own index key, because most of them need different things stored so they can be converted to rows efficiently after getting the key from the index.
- There is some common logic to find packfiles for a given repository and get from there the offsets of the objects that will be indexed.
- For `ref_commits`, `commit_trees` and `commit_blobs` there is a common logic `rowKeyValueIter` and `rowIndexIter`, since they are stored as is. The other tables had no obvious common logic, so instead of coming up with an abstraction that might be a PITA in the future, I preferred to have some more tedious and perhaps slightly repeating implementations on the other iterators (though there are no exact identical impls).


ajnavarro:
Related: https://github.com/src-d/go-git/pull/854

commits.go
```diff
@@ -261,3 +294,107 @@ func (i *commitsByHashIter) nextList() (*object.Commit, error) {
 		return commit, nil
 	}
 }
+
+type commitIndexKey struct {
```
ajnavarro: Could we have a generic `ObjectIndexKey` struct for all the objects that are into the packfile?

erizocosmico:
Only commit and blob have the struct totally in common. Tree entries also needs the pos inside the tree and the rest of the tables are very different. I can make one common object for blobs and commits tho.
On Tue, 5 Jun 2018 at 18:51, Antonio Navarro Perez <notifications@github.com<mailto:notifications@github.com>> wrote:

@ajnavarro commented on this pull request.

________________________________

In commits.go<https://github.com/src-d/gitbase/pull/298#discussion_r193145645>:

> @@ -261,3 +294,107 @@ func (i *commitsByHashIter) nextList() (*object.Commit, error) {
                return commit, nil
        }
 }
+
+type commitIndexKey struct {


Could we have a generic ObjectIndexKey struct for all the objects that are into the packfile?

erizocosmico:
Added the tests and some fixes, along with the unification of the index keys where it was possible. It can be reviewed now.

ajnavarro:
go-git 4.4.1 has dotgit logic outside `internal` folder, so we can use it now updating the dependency.

erizocosmico:
@ajnavarro Since this is against a feature branch and won't go straight to master, what do you think if we review and merge this one first and then in a followup PR I update the dependency and replace the manual impl with the dotgit one?

packfiles.go
```diff
@@ -199,15 +199,15 @@ func (i *objectIter) Next() (*encodedObject, error) {
...
 			RepositoryID: i.packs.repo.path,
-			Packfile:     i.packs.packfiles[i.packs.packpos],
+			Packfile:     i.packs.packfiles[i.packs.packpos-1],
 		}, nil
 	}
 }
 
 func (i *objectIter) Close() error {
-	if i.packObjects != nil {
+	/*if i.packObjects != nil {
```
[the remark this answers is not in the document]
erizocosmico: yep

erizocosmico:
Fixed errors in travis

cmd/gitbase/command/server.go
```diff
@@ -28,13 +32,15 @@ const (
...
-	Password string   `short:"P" long:"password" default:"" description:"Password used for connection"`
+	Verbose   bool     `short:"v" description:"Activates the verbose mode"`
+	Git       []string `short:"g" long:"git" description:"Path where the git repositories are located, multiple directories can be defined. Accepts globs."`
+	Siva      []string `long:"siva" description:"Path where the siva repositories are located, multiple directories can be defined. Accepts globs."`
+	Host      string   `short:"h" long:"host" default:"localhost" description:"Host where the server is going to listen"`
+	Port      int      `short:"p" long:"port" default:"3306" description:"Port where the server is going to listen"`
+	User      string   `short:"u" long:"user" default:"root" description:"User name used for connection"`
+	Password  string   `short:"P" long:"password" default:"" description:"Password used for connection"`
+	PilosaURL string   `long:"pilosa" default:"http://localhost:10101" description:"URL to your pilosa server"`
+	IndexDir  string   `short:"i" long:"index" default:"/var/gitbase/index" description:"Directory where the gitbase indexes information will be persisted."`
```
ajnavarro: Maybe we should change the default path. What do you think @jfontan ?

cmd/gitbase/command/server.go
```diff
@@ -104,6 +110,18 @@ func (c *Server) buildDatabase() error {
 	c.engine.Catalog.RegisterFunctions(function.Functions)
 	logrus.Debug("registered all available functions in catalog")
 
+	if err := os.MkdirAll(c.IndexDir, 0755); err != nil {
```
ajnavarro: Should we add some log on these new steps?

files.go
```diff
@@ -251,3 +299,195 @@ func fileToRow(
...
+		values, err := rowIndexValues(row, i.columns, FilesSchema)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		return values, key, nil
+	}
+}
+
+func (i *filesKeyValueIter) Close() error {
```
ajnavarro: We should call this on tests too.

repositories.go
```diff
@@ -93,6 +94,46 @@ func (r *repositoriesTable) WithProjectAndFilters(
...
+	span, ctx := ctx.Span("gitbase.RepositoriesTable.WithProjectFiltersAndIndex")
+	s, ok := ctx.Session.(*Session)
+	if !ok || s == nil {
+		span.Finish()
+		return nil, ErrInvalidGitbaseSession.New(ctx.Session)
+	}
+
+	var iter sql.RowIter = &rowIndexIter{index}
+
+	if len(filters) > 0 {
```
ajnavarro: Can we test this with some filter?

ajnavarro:
Check codecov, appears that squash_join is not tested at all.

cmd/gitbase/command/server.go
erizocosmico: What path should we change it to?

jfontan: Usually software does not have var directories directly in `/var` but in one of its subdirectories. MySQL has its database files in `/var/lib/mysql`.

erizocosmico:
Fixed the coverage thing of the squash tables

remotes.go
```diff
@@ -99,6 +101,46 @@ func (r *remotesTable) WithProjectAndFilters(
...
+	return &remotesKeyValueIter{repos: iter, columns: colNames}, nil
+}
+
+// WithProjectFiltersAndIndex implements sql.Indexable interface.
+func (*remotesTable) WithProjectFiltersAndIndex(
+	ctx *sql.Context,
+	columns, filters []sql.Expression,
+	index sql.IndexValueIter,
+) (sql.RowIter, error) {
+	span, ctx := ctx.Span("gitbase.ReferencesTable.WithProjectFiltersAndIndex")
```
jfontan: Shouldn't this be `"gitbase.RemotesTable.WithProjectFiltersAndIndex"`?

erizocosmico: Fixed

erizocosmico:
Everything is fixed now, upgraded go-mysql-server and added engine initialization to the server command.

PR #308, opened 2018-06-13
gitbase: support for unpacked objects, integration tests added

erizocosmico:
This PR adds support for unpacked objects, which made the index contain different objects and not match the result of the query without indexes.
A bunch of integration tests for indexes of all tables testing they output the exact same thing as the same query without indexes have been added as well.

Because it's super important to ensure the index is closed so it can be released, each index iterator has now a call to `closeIndexOnError`, which closes the index if there is an error or EOF in an iterator, ensure it's *always* closed.

After this is merged, `feature/indexable` is ready to be PR'd and merged to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,11 @@ addons:
     - gcc-6
     - g++-6
 
+before_install:
+  - docker pull pilosa/pilosa:v0.9.0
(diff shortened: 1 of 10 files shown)


erizocosmico:
Added start of pilosa container to the .travis.yml, so build should be fine now

PR #310, opened 2018-06-14
Implement indexes

erizocosmico:
Closes #253 

We should be able to release a version with indexes after this.

PR #311, opened 2018-06-18
Do not use full path in plain repository IDs

### LINKED ISSUE #307: Use relative paths to repositories on IndexValues. Throw a warning when two repo identifiers are the same using the -g command line option.

jfontan:
The base path that is stripped to create the ID found as follows:

* If the path does not contain glob chars all of it
* If it contains glob pattern all path components previous to the first
pattern occurrence

This way we can add bucketed directories and have the repositories at
the same level. For example, adding "bucket/*":

    bucket/a/aa.git -> aa.git
    bucket/a/ab.git -> ab.git
    bucket/b/ba.git -> ba.git
    bucket/b/bb.git -> bb.git

Now a repository is only added once and an error log is printed if a new
repo with the same ID tries to be registered. This allows adding
complementary repo. Example, add new and old download dirs.

**Note**: it uses the number of path components as prefix parameter to `AddDir` instead of the prefix itself like the siva version as it seems more logical to me in this context. This can be changed to use a string prefix for coherence with the siva version.

Fixes #307


erizocosmico:
Can you try with #310 to see if this will break the indexes? right now, a lot relies on the repository_id to open the correct packfile, perhaps this will needs some tweaks to work with that

jfontan:
Index tests use siva files which behavior has not been changed. I'll add standard git repos to the test.

jfontan:
Changed the tests in integration to use no related repository IDs and still work with these changes.

erizocosmico:
You need to rebase

jfontan:
Rebased with master and changed integration test to use a repository ID != path.

repository_pool.go
```diff
@@ -100,41 +106,59 @@ func NewRepositoryPool() *RepositoryPool {
...
+	return p.AddGitWithID(path, path)
+}
+
+// AddGitWithID checks if a git repository can be opened and adds it to the
+// pool. ID should be specified.
+func (p *RepositoryPool) AddGitWithID(id, path string) (string, error) {
 	_, err := git.PlainOpen(path)
 	if err != nil {
-		return "", err
+		return "", errRepoCanNotOpen.New(err)
```
ajnavarro: Should we use `Wrap(err, path)` instead of `New` here?

util.go
```diff
@@ -0,0 +1,73 @@
+package gitbase
```
ajnavarro: Can we change the name of that file to `path_utils.go` or something like that? I would prefer specific utils that all-in-one utils file with a lot of crap on it.

jfontan:
Added path to `errRepoCannotOpen` error.

PR #312, opened 2018-06-18
*: add a comment in the README pointing to the supported SQL subset

### LINKED ISSUE #306: Document which parts of SQL are supported
During my friction log at #303 I could not find the documentation for what features of SQL are supported.

This issue is to track the creation of either a static documentation page, or a subcommand in `gitbase` that will show the verbs that are (not) supported.

mcarmonaa:
Closes #306 

Depends on https://github.com/src-d/go-mysql-server/pull/226

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,10 @@ thousands of repositories in a single node. Stay tuned!
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,10 @@ thousands of repositories in a single node. Stay tuned!
 
 ## Examples
 
+To see the SQL subset currently supported take a look at [this list](https://github.com/src-d/go-mysql-server/blob/master/SUPPORTED.md) from [src-d/go-mysql-server](https://github.com/src-d/go-mysql-server).
```
ajnavarro: Should we add a link to the same commit as is used by the project? to avoid misunderstandings between the actual go-mysql-version and the go-mysql-version that is in use on gitbase.

mcarmonaa: Sure

PR #339, opened 2018-06-26
gitbase: ignore .DS_Store files in darwin

erizocosmico:
Closes #337

PR #341, opened 2018-06-27
travis: build and upload macOS releases

### LINKED ISSUE #328: Make Darwin compilation works
Due to some cgo dependencies, we cannot cross-compile Darwin binaries. Compile them using a Darwin Travis node. Bblfsh cannot run on Darwin, so we will need to skip integration tests doing like here: https://github.com/bblfsh/client-go/blob/64416eecb5321d57179f787750f191fb6da045ce/client_test.go#L25

erizocosmico:
Closes #328 

Since our codebase does not have anything that is os-dependant on the darwin build I'm just building the binary just to know that it can be built in macOS with no problems.

Right now, you can't really start a docker container inside a macOS build, so there is no way we can run bblfsh there and we would need to reduce the number of tests we run for macOS quite a bit using the `-short` hack, which is kind of a pointless effort.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,4 @@
 language: go
-sudo: required
-
-go:
-  - "1.9.x"
-  - "1.10.x"
-  - tip
-
-services:
-  - docker
 
 go_import_path: github.com/src-d/gitbase
 
@@ -28,10 +19,10 @@ before_install:
   - docker pull pilosa/pilosa:v0.9.0
   - docker run -d --name pilosa -p 127.0.0.1:10101:10101 pilosa/pilosa:v0.9.0
   - docker ps -a
-
-install:
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 90
   - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-6 90
+
+install:
   - cd ./vendor/gopkg.in/bblfsh/client-go.v2
   - make dependencies
   - cd ./../../../..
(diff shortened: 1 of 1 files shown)


ajnavarro:
should we ignore all hidden files and folders?

erizocosmico:
What do you mean by hidden files and folders?

ajnavarro:
Sorry I mix that issue with #339

.travis.yml
```diff
@@ -45,17 +36,55 @@ before_script:
...
-deploy:
-  provider: releases
-  api_key: $GITHUB_TOKEN
-  file_glob: true
-  file: build/*.tar.gz
-  skip_cleanup: true
-  on:
-    tags: true
+      script:
+        - make packages || echo "" # will fail because of docker being missing
```
smola: I think make packages will print an error if Docker is missing, but will not fail.

.travis.yml
```diff
@@ -45,17 +36,55 @@ before_script:
...
+      deploy:
+        provider: releases
+        api_key: $GITHUB_TOKEN
+        file_glob: true
+        file: build/*linux_amd64.tar.gz
+        skip_cleanup: true
+        on:
+          tags: true
+
+    - {go: 1.9.x, os: linux, sudo: required, dist: trusty, services: [docker]}
```
jfontan: We can skip 1.9.

erizocosmico:
Done @jfontan

.travis.yml
ajnavarro: @erizocosmico ^^^

erizocosmico: Yes, it does fail, I had to add `|| echo ""` because it did make it fail :( (it would be `|| true`, but macOS travis does not have `true`)

PR #336, opened 2018-06-26
*: implement commit_files table

erizocosmico:
Implements `commit_files` table with squash and indexes.

PR #344, opened 2018-06-28
gitbase: custom encode and decode of index keys to save space

erizocosmico:
By using custom encoder and decoder for the index keys, as opposed to golang `gob` we can save a lot of space.

This implements our own way of encoding and decoding those index keys by just writing bytes for each thing.

- For string values we write first the size (int64 in binary) and then the string itself.
- For hashes we just write 40 bytes.
- For int64 we write the number in binary.
- For bools we write 0 or 1.

Since some objects can be unpacked and not in the packfile, there are cases where the key has an offset of -1 and a hash instead or an offset >= 0 and empty hash.

Because now we're in control, instead of saving both the hash and the offset we store a bool and just one of them, potentially saving some space. Also, since this is not using reflection it should also be way faster than before.

By definition, this takes less space than before and this is true no matter which key you're encoding because gob stores more than just the value and we do not. But just to have some numbers I did an index on `files`, which is the most space consuming table, saving `file_path` and `blob_hash`.

I did it on a repository of 113MB with 197500 rows (src-d/datasets).
- Previous index size was 109.1MB
- Current index size was 75MB
In this case, we saved 31.25% of space, so I'd say it's ok.

### DIFF
--- a/files.go
+++ b/files.go
@@ -1,6 +1,7 @@
 package gitbase
 
 import (
+	"bytes"
 	"io"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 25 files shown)


erizocosmico:
# Don't merge, needs changes after merging #336

erizocosmico:
It's been updated for `commit_files` and can be reviewed

index.go
```diff
@@ -102,17 +113,143 @@ func (i *rowIndexIter) Next() (sql.Row, error) {
...
+	b, err := buf.ReadByte()
+	if err != nil {
+		return false, fmt.Errorf("can't read bool: %s", err)
+	}
+
+	return b == 1, nil
+}
+
+func writeInt64(buf *bytes.Buffer, n int64) {
+	var bs = make([]byte, 8)
```
jfontan: Do we want to use a `Varint` or a full size `Int64`? From the documentation the size of a `Varint` can be 10 bytes:

https://golang.org/src/encoding/binary/varint.go

erizocosmico: You'e right. I will change it to how it was before, storing a `Uint64`

erizocosmico: Done

files.go
```diff
@@ -298,6 +299,75 @@ type fileIndexKey struct {
...
+			return err
+		}
+		k.Offset = -1
+	}
+
+	if k.Name, err = readString(buf); err != nil {
+		return err
+	}
+
+	if k.Mode, err = readInt64(buf); err != nil {
```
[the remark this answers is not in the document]
erizocosmico: Yep, it was just to not add a one more type of encode/decode just for a single value

index.go
```diff
@@ -102,17 +113,149 @@ func (i *rowIndexIter) Next() (sql.Row, error) {
...
+	bs := []byte(s)
+	writeInt64(buf, int64(len(bs)))
+	buf.Write(bs)
+}
+
+var errInvalidHashSize = errors.NewKind("invalid hash size: %d, expecting 40 bytes")
+
+func writeHash(buf *bytes.Buffer, s string) error {
+	bs := []byte(s)
+	if len(bs) != 40 {
```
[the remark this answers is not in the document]
erizocosmico: Done

PR #356, opened 2018-06-29
cmd/gitbase: add readonly flag

### LINKED ISSUE #326: Restrict CREATE INDEX queries
Now that we have `CREATE INDEX` it might be a good idea having a way to restrict who is allowed to perform it. For example, if we offered a gitbase-playground service open to the public.

Full user permission management might be overkill, but maybe we could have 2 default users for full-access and read-only access; `root` and `user`.

erizocosmico:
Closes #326 

Depends on https://github.com/src-d/go-mysql-server/pull/251, so build will fail until go-mysql-server is upgraded to that

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -485,7 +485,11 @@
     "sql/parse",
     "sql/plan"
   ]
+<<<<<<< HEAD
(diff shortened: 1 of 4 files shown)


erizocosmico:
Updated go-mysql-server, CI should work now

ajnavarro:
@erizocosmico yes please.

PR #358, opened 2018-06-29
cmd/server: add trace parameters to enable jaeger tracing

jfontan:
I have not added examples in documentation as tracing does not seem a user facing feature. CLI help is updated.


cmd/gitbase/command/server.go
```diff
@@ -39,9 +43,10 @@ type Server struct {
 	PilosaURL     string   `long:"pilosa" default:"http://localhost:10101" description:"URL to your pilosa server" env:"PILOSA_ENDPOINT"`
 	IndexDir      string   `short:"i" long:"index" default:"/var/lib/gitbase/index" description:"Directory where the gitbase indexes information will be persisted." env:"GITBASE_INDEX_DIR"`
 	DisableSquash bool     `long:"no-squash" description:"Disables the table squashing."`
-	// IgnoreGitErrors by default when gitbase encounters and error in a
-	// repository it stops the query. With this parameter it won't complain and
-	// just skip those rows or repositories.
+	TraceEnabled  bool     `long:"trace" env:"GITBASE_TRACE" description:"Enables jaeger tracing"`
+	TraceEndpoint string   `long:"trace-endpoint" env:"GITBASE_TRACE_ENDPOINT" default:"localhost:6831" description:"Sets endpoint to send traces"`
```
ajnavarro: There are some jaeger standard env vars that we can use instead of this:https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-core#configuration-via-environment

docs/using-gitbase/configuration.md
```diff
@@ -41,14 +41,18 @@ Help Options:
...
-      -g, --git=      Path where the git repositories are located, multiple directories can be defined. Accepts globs.
-          --siva=     Path where the siva repositories are located, multiple directories can be defined. Accepts globs.
-      -h, --host=     Host where the server is going to listen (default: localhost)
-      -p, --port=     Port where the server is going to listen (default: 3306)
-      -u, --user=     User name used for connection (default: root)
-      -P, --password= Password used for connection
-          --pilosa=   URL to your pilosa server (default: http://localhost:10101)
-      -i, --index=    Directory where the gitbase indexes information will be persisted. (default: /var/lib/gitbase/index)
-          --no-squash Disables the table squashing.
+      -v                    Activates the verbose mode
```
ajnavarro: At the top of this file, we have a table with several env variables. Maybe we should add the new ones there too.

jfontan:
Now only `--trace` is added to server parameters. The rest is handled by jaeger environment variables. I've also added logrus logging to jaeger for easier debugging.

The table of jaeger environment variables is copied with the same format so it is easier to update if jaeger adds or changes them.

cmd/gitbase/command/server.go
```diff
@@ -66,12 +79,54 @@ func (c *Server) Execute(args []string) error {
 		{Password: c.Password},
 	}
 
+	var tracer opentracing.Tracer
+	if c.TraceEnabled {
+		if os.Getenv("JAEGER_SERVICE_NAME") == "" {
```
ajnavarro: Maybe instead of checking an environment variable, will be less error prone check the content of https://sourcegraph.com/github.com/src-d/gitbase@8844a396d235d32143f77bf85ec9a1820bbf538d/-/blob/vendor/github.com/uber/jaeger-client-go/config/config.go#L38 into the cfg object.

cmd/gitbase/command/server.go
```diff
@@ -66,12 +79,54 @@ func (c *Server) Execute(args []string) error {
...
+
+		if err != nil {
+			logrus.WithField("error", err).Fatal("unable to initialize tracer")
+			return err
+		}
+
+		tracer = t
+		defer closer.Close()
+
+		jaegerHost := os.Getenv("JAEGER_AGENT_HOST")
```
ajnavarro: the same as the previous comment.

cmd/gitbase/command/server.go
```diff
@@ -66,12 +79,54 @@ func (c *Server) Execute(args []string) error {
...
+		if jaegerHost == "" {
+			jaegerHost = jaeger.DefaultUDPSpanServerHost
+		}
+
+		jaegerPort := os.Getenv("JAEGER_AGENT_PORT")
```
ajnavarro: the same as the previous comment.

docs/using-gitbase/configuration.md
```diff
@@ -27,28 +50,33 @@ Available commands:
...
 
-```bash
+```
 Usage:
   gitbase [OPTIONS] server [server-OPTIONS]
 
 Starts a gitbase server instance
 
-By default when gitbase encounters an error in a repository it
+By default when gitbase encounters and error in a repository it
```
ajnavarro: I think this is `an` not `and`

cmd/gitbase/command/server.go
```diff
@@ -39,16 +44,24 @@ type Server struct {
 	PilosaURL     string   `long:"pilosa" default:"http://localhost:10101" description:"URL to your pilosa server" env:"PILOSA_ENDPOINT"`
 	IndexDir      string   `short:"i" long:"index" default:"/var/lib/gitbase/index" description:"Directory where the gitbase indexes information will be persisted." env:"GITBASE_INDEX_DIR"`
 	DisableSquash bool     `long:"no-squash" description:"Disables the table squashing."`
-	// IgnoreGitErrors by default when gitbase encounters and error in a
-	// repository it stops the query. With this parameter it won't complain and
-	// just skip those rows or repositories.
+	TraceEnabled  bool     `long:"trace" env:"GITBASE_TRACE" description:"Enables jaeger tracing"`
```
ajnavarro: You have a default env variable to do this: `JAEGER_DISABLED`. I don't know the default value. We should check that. Maybe the default tracer with no env variables is a `noop tracer`

cmd/gitbase/command/server.go
jfontan: The configuration is still not created at that point. Also, setting the environment is what fills `ServiceName` in `FromEnv`. That's how they do in the examples:

https://github.com/jaegertracing/jaeger-client-go/blob/master/config/example_test.go#L107-L110

cmd/gitbase/command/server.go
jfontan: I was not able to get those parameters from config. These are set in UDP transport and they don't have any public way to get them.

https://github.com/jaegertracing/jaeger-client-go/blob/master/transport_udp.go#L49-L75

cmd/gitbase/command/server.go
jfontan: By default it is enabled and the tracer is jaeger. That's what happens now when you specify `--tracer` and no env vars. I preferred tracing to be completely disabled by default.

docs/using-gitbase/configuration.md
jfontan: Fixed

cmd/gitbase/command/server.go
jfontan: Changed :+1:

cmd/gitbase/command/server.go
jfontan: Deleted generation of the endpoint in logs as it's not really needed.

cmd/gitbase/command/server.go
jfontan: Deleted

jfontan:
I'll rebase after review is finished.

ajnavarro:
rebase and LGTM!

jfontan:
@ajnavarro rebase done

PR #371, opened 2018-07-04
New rev-upgrade script

kuba--:

No ruby, no python, just go.


pull_request_template.md
```diff
@@ -5,6 +5,6 @@ All PRs must keep the documentation up to date. If this PR changes or adds some
 - Schema changes
 - Syntax changes
 - Add or update examples
-- ./rev-upgrade.sh "gopkg.in/src-d/go-mysql-server.v0" "the latest go-mysql-server revision"
+- `go run ./cmd/rev-upgrade/main.go -p "gopkg.in/src-d/go-mysql-server.v0" [-r "revision"]`
```
erizocosmico: The `-p` could be optional as well, since this is mainly for updating the rev of go-mysql-server on the docs

cmd/rev-upgrade/main.go
```diff
@@ -0,0 +1,187 @@
...
+	var (
+		prj    string
+		newRev string
+		oldRev string
+
+		w   *git.Worktree
+		err error
+	)
+
+	flag.StringVar(&prj, "p", "", "project name (e.g.: gopkg.in/src-d/go-mysql-server.v0)")
```
erizocosmico: maybe we could make this optional and make `gopkg.in/src-d/go-mysql-server.v0` the default, since it's the main (and only for now) use of the script

ajnavarro:
I would propose something simpler:

To change go-mysql-server dependency we need to change Gopkg.toml file and do a dep ensure for sure.

We can add a makefile command to check if the updated hash appears more than one time on the repository data. If not, throw an error on Travis.

Doing that, we don't need to remember to execute the command, or have it prepared to run in several OS.

I think it is a more KISS solution, WDYT guys?

kuba--:
@erizocosmico - I also wondered where to put the script, and finally I've moved it to `tools/rev-upgrade/main.go`. Also made `-p` optional.
@ajnavarro - I would go with any _KISS_ solution (this is what I quickly wrote for myself in awk), but because I decided to invest some time in that, so I made something generic (totally not related to gitbase, mysql, version, etc.) Actually it can be used anywhere.
Anyway, it's for us. If it's not useful or over complicated, we can throw it away (have it as a `Makefile` target is not a bad idea).

PR #372, opened 2018-07-04
internal/function: don't fail if the uast function can't parse a blob

### LINKED ISSUE #369: Do not fail a query for bblfsh error in one row
Right now if I run a query asking for a `UAST()` column, the whole query returns an error if bblfsh fails to parse any of the rows.

This is inconvenient because a file with a syntax error can prevent any query from working for all the other files.

```
MySQL [(none)]> select UAST(blob_content, LANGUAGE(file_path, blob_content)) from files where repository_id="demo";
ERROR 1105 (HY000): unknown error: unable to parse the given blob using bblfsh: Syntax error on token "publi", public expected
```

I think there should be a way to get all the rows, with an empty UAST value where bblfsh failed.

What would be awesome is to have these bblfsh errors for each failed UAST request. Maybe there is a way to make `UAST()` return either UAST, or error messages?

mcarmonaa:
Fixes #369

### DIFF
--- a/internal/function/uast.go
+++ b/internal/function/uast.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"github.com/src-d/gitbase"
 	"gopkg.in/bblfsh/client-go.v2/tools"
 	"gopkg.in/bblfsh/sdk.v1/protocol"
@@ -197,11 +198,12 @@ func (f UAST) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 
 	resp, err := client.Parse(ctx, lang, bytes)
 	if err != nil {
-		return nil, ErrParseBlob.New(err)
+		logrus.Warn(ErrParseBlob.New(err))
+		return nil, nil
 	}
 
 	if resp.Status != protocol.Ok {
-		return nil, ErrParseBlob.New(strings.Join(resp.Errors, "\n"))
+		logrus.Warn(ErrParseBlob.New(strings.Join(resp.Errors, "\n")))
 	}
 
 	var nodes []*uast.Node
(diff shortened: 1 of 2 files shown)


internal/function/uast.go
```diff
@@ -197,11 +198,13 @@ func (f UAST) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 
 	resp, err := client.Parse(ctx, lang, bytes)
 	if err != nil {
-		return nil, ErrParseBlob.New(err)
+		logrus.Warn(ErrParseBlob.New(err))
```
ajnavarro: An error can be thrown an the protocol result be OK

ajnavarro: So, we should check if resp is nil too to fail here (I think).

mcarmonaa: the protocol result shouldn't be `OK` if there is an error and if resp is not nil it's because the status if `FATAL`, so we must return any case

```
// DoWithContext does the same as Do(), but sopporting cancellation by the use
// of Go contexts.
func (r *ParseRequest) DoWithContext(ctx context.Context) (*protocol.ParseResponse, error) {
	if r.err != nil {
		return nil, r.err
	}

	resp, err := r.client.service.Parse(ctx, &r.internal)
	if err != nil {
		return nil, err
	} else if resp.Status == protocol.Fatal {
		return resp, FatalError(resp.Errors)
	}
	return resp, nil
}
```

ajnavarro: yep, you are right. But in the case that the error is not fatal, we should print the errors from the response as warnings and return the partial UAST.

internal/function/uast.go
```diff
@@ -197,11 +198,13 @@ func (f UAST) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
...
+		return nil, nil
 	}
 
 	if resp.Status != protocol.Ok {
```
ajnavarro: I mean here.

mcarmonaa: done

PR #374, opened 2018-07-04
mysql-upgrade target

kuba--:

command `make mysql-upgrade` upgrades mysql to the latest (master) revision

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ CI_REPOSITORY ?= https://github.com/src-d/ci.git
 CI_PATH ?= $(shell pwd)/.ci
 CI_VERSION ?= v1
 
+UPGRADE_PRJ ?= "gopkg.in/src-d/go-mysql-server.v0"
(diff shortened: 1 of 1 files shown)


kuba--:
whatever just something what we can easily remember (like we have dependencies :))

kuba--:
we may also have sth. more generic:
```
make PRJ="gopkg.in/src-d/go-mysql-server.v0" upgrade
```

erizocosmico:
That IMHO is a little bit verbose

kuba--:
we may always have something short:
`make upgrade`
and if some day we will have to upgrade any other project then we may split this target into two or add another `go run tools/rev-upgrade/main.go` under the same `upgrade` target.

ajnavarro:
@kuba-- go with `make upgrade`

kuba--:
OK, I've changed it to: `make upgrade`.

By default it upgrades mysql, but if we want to apply upgrade to any different project we can do it by: 
```
make UPGRADE_PRJ="project name" UPGRADE_REV="revision number" upgrade
```

PR #375, opened 2018-07-05
docs: add indexes section

mcarmonaa:
Signed-off-by: Manuel Carmona <manu.carmona90@gmail.com>

### DIFF
--- a/docs/using-gitbase/indexes.md
+++ b/docs/using-gitbase/indexes.md
@@ -1,0 +1,13 @@
+# Indexes
+
+`gitbase` allows you to speed up queries creating indexes.
+
+Indexes are implemented as bitmaps using [pilosa](https://github.com/pilosa/pilosa) as a backend storage for them. To run a pilosa instance see the [getting started](./getting-started.md) section.
+
+Thus, to create indexes you must specify pilosa as the type of index. You can find some examples in the [examples](./examples.md#create-an-index-for-columns-on-a-table) section about managing indexes.
+
+Note that you can create an index either **on one or more columns** or **on a single expression**.
+
+You can find some more examples in the [examples](./examples.md#create-an-index-for-columns-on-a-table) section.
+
+See [go-mysql-server](https://github.com/src-d/go-mysql-server/tree/e03c0c8264d306f683a8c1e7933c5d2f4b200a5b#indexes) documentation for more details


docs/using-gitbase/indexes.md
```diff
@@ -0,0 +1,16 @@
...
+Thus, to create indexes you must specify pilosa as the type of index.
+```
+CREATE INDEX files_path_idx ON files USING pilosa (file_path)
+```
+
+Note that you can create an index either **on one or more columns** or **on a single expression**.
+
+You can find some more examples in the [examples](./examples.md) section.
+
+See [go-mysql-server](https://github.com/src-d/go-mysql-server) documentation for more details
```
erizocosmico: We should link to the specific version, as it's done in some other parts of the docs

mcarmonaa: Done, note that for the moment this link doesn't take you to the `Indexes` section in the `README` because there's no such a section in this revision. When https://github.com/src-d/go-mysql-server/pull/264 will be merged we will update the`go-mysql-server` dependency and get the correct link.

docs/using-gitbase/indexes.md
```diff
@@ -0,0 +1,16 @@
+# Indexes
+
+`gitbase` allows you to speed up queries creating indexes.
+
+Indexes are implemented as bitmaps using [pilosa](https://github.com/pilosa/pilosa) as a backend storage for them. To run a pilosa instance see the [getting started](./getting-started.md) section.
+
+Thus, to create indexes you must specify pilosa as the type of index.
+```
+CREATE INDEX files_path_idx ON files USING pilosa (file_path)
```
ajnavarro: I would add some example of using more than one column and using an expression.

mcarmonaa: There are already examples for that in `examples.md` and a link to it. Should I add them here anyway?

mcarmonaa: I added the specific section in `examples.md` to the link, btw

ajnavarro: I would say if we have an example here, complete it with all the possible index cases.

ajnavarro: If it is not complete, I would say to delete it to do not have duplicate info.

mcarmonaa: So I'll remove it then and place here the link to the examples.

PR #382, opened 2018-07-11
gitbase: compress keys using zlib

### LINKED ISSUE #381: Key too large saving indexes for files

erizocosmico:
Closes #381 (it should)

On very large keys (such as files), the space is almost ~50% of what it is without compressing, so that's good.

It will be slower, though.

### DIFF
--- a/index.go
+++ b/index.go
@@ -2,10 +2,12 @@ package gitbase
 
 import (
 	"bytes"
+	"compress/zlib"
 	"crypto/sha1"
 	"encoding/binary"
 	"fmt"
 	"io"
+	"io/ioutil"
 
 	errors "gopkg.in/src-d/go-errors.v1"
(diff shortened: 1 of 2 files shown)


index.go
```diff
@@ -25,11 +27,36 @@ type Indexable interface {
...
 
 func encodeIndexKey(k indexKey) ([]byte, error) {
-	return k.encode()
+	bs, err := k.encode()
+	if err != nil {
+		return nil, err
+	}
+
+	var buf bytes.Buffer
+	gz := zlib.NewWriter(&buf)
```
ajnavarro: Zlib Writer creation is really expensive. You should reuse the Writer calling to `Writer.Reset()`.

erizocosmico:
Done @ajnavarro

PR #384, opened 2018-07-11
gitbase/repository_pool: change repository to be an interface

jfontan:
Also add `billyRepo` for billy filesystems.

I'm not totally happy with the naming of `repository` interface as there's already a `Repository` structure and all the `gitRepository`/`gitRepo` for struct and constructor. Suggestions welcome.

### DIFF
--- a/fs_error_test.go
+++ b/fs_error_test.go
@@ -60,15 +60,15 @@ func setupErrorRepos(t *testing.T) (*sql.Context, CleanupFunc) {
 
 	fs, err := brokenFS(brokenPackfile, baseFS)
 	require.NoError(err)
-	pool.AddBilly("packfile", fs)
+	pool.Add(billyRepo("packfile", fs))
 
 	fs, err = brokenFS(brokenIndex, baseFS)
 	require.NoError(err)
-	pool.AddBilly("index", fs)
+	pool.Add(billyRepo("index", fs))
 
 	fs, err = brokenFS(0, baseFS)
 	require.NoError(err)
-	pool.AddBilly("ok", fs)
+	pool.Add(billyRepo("ok", fs))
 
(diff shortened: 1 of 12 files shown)


erizocosmico:
Why do we need to add billy repositories to gitbase? IMHO, testing some more filesystem errors might not be worth introducing all this exclusively for testing.

jfontan:
@erizocosmico moved `billyRepository` to the test file (`fs_error_test.go`)

fs_error_test.go
```diff
@@ -81,6 +81,25 @@ func setupErrorRepos(t *testing.T) (*sql.Context, CleanupFunc) {
...
+	brokenType brokenType,
+	fs billy.Filesystem,
+) (billy.Filesystem, error) {
+	dotFS, err := fs.Chroot(".git")
+	if err != nil {
+		return nil, err
+	}
+
+	var brokenFS billy.Filesystem
+	if brokenType == 0 {
```
kuba--: what's _0_? do we need a const for this?

erizocosmico:
What I mean is: do we really benefit from making Repository an interface and the complexity this comes with just to test some FS errors?

jfontan:
Even if we don't add more types of repositories and move to other ways to test errors (maybe fixtures with broken packfiles) I still think it makes sense. It puts the different logic for siva and git repos in one place instead of different code paths in other parts:

https://github.com/src-d/gitbase/pull/384/files#diff-b7e4ef15e429a4c21afc4a8822b635beL129

fs_error_test.go
jfontan: It means no broken flags, I've just added a constant.

jfontan:
@ajnavarro I've just rebased

PR #410, opened 2018-07-31
show indexes support

mcarmonaa:
This PR just update the go-mysql-server dependency to make the use of `show indexes` available.

Note that the only valuable information displayed for the moment are the name of the indexes and the type of the index for the requested table. 

```
MySQL [(none)]> show indexes from refs;
+-------+------------+---------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+------------+
| Table | Non_unique | Key_name      | Seq_in_index | Column_name | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment | Visible | Expression |
+-------+------------+---------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+------------+
| refs  |          1 | refs_name_idx |            0 | NULL        |           |           0 |        0 |        |      | pilosa     |         |               | YES     | NULL       |
+-------+------------+---------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+------------+
1 row in set (0.00 sec)

```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,47 +2,36 @@
 
 
 [[projects]]
-  digest = "1:08108caf6ec616d19308408dbd873961bdbf32dc10694d8824d162cce026dc90"
   name = "github.com/BurntSushi/toml"
   packages = ["."]
-  pruneopts = ""
   revision = "b26d9c308763d68093482582cea63d69be07a0f0"
   version = "v0.3.0"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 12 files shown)


mcarmonaa:
Sorry, I thought I had done it.

Doesn't `make upgrade` do the same?

mcarmonaa:
Ok, we had merged a previous dependency update without run 'make upgrade', so when I ran it before it didn't match any file because they didn't contain the previous revision from `Gopkg.toml`. Those files that should have matched contained an older revision.

Now is updated properly @ajnavarro.

PR #416, opened 2018-08-08
cmd/gitbase: load siva files and git indistinctly

### LINKED ISSUE #414: Automatically detect siva/git
Instead of having flags for `git` and `siva` have it automatically detect the types in the subtree of the folder (maybe with a depth flag).

mcarmonaa:
Closes #414 

The new usage looks like:
```sh
Usage:
  gitbase [OPTIONS] server [server-OPTIONS]

Starts a gitbase server instance

By default when gitbase encounters an error in a repository it
stops the query. With GITBASE_SKIP_GIT_ERRORS variable it won't
complain and just skip those rows or repositories.

Help Options:
  -h, --help             Show this help message

[server command options]
      -v                 Activates the verbose mode
      -d, --directories= Path where the git repositories are located (standard and siva), multiple directories can be defined. Accepts globs.
          --depth=       load repositories looking at less than <depth> nested subdirectories. (default: 1000)
          --no-git       disable the load of git standard repositories.
          --no-siva      disable the load of siva files.
          --host=        Host where the server is going to listen (default: localhost)
      -p, --port=        Port where the server is going to listen (default: 3306)
      -u, --user=        User name used for connection (default: root)
      -P, --password=    Password used for connection
          --pilosa=      URL to your pilosa server (default: http://localhost:10101) [$PILOSA_ENDPOINT]
      -i, --index=       Directory where the gitbase indexes information will be persisted. (default: /var/lib/gitbase/index) [$GITBASE_INDEX_DIR]
          --no-squash    Disables the table squashing.
          --trace        Enables jaeger tracing [$GITBASE_TRACE]
      -r, --readonly     Only allow read queries. This disables creating and deleting indexes as well. [$GITBASE_READONLY]

```

***Changes***
- the flag to specify a list of directories now is `-d, --directories` so there's no more `--git` or `--siva`
> gitbase server -d=/d1,/d2

- The directories are traversed recursively, but with the new flag `--depth` you can specify how many levels from the specified directory must be checked.
> gitbase server -d=/d1,/d2 --depth 1

- Now siva files and git repositories are loaded even if the they are mixed so if you don't want to load some of them you can use the flags `--no-git` and `--no-siva`:
> gitbase server -d=/d1,/d2 --no-siva

So if we have a directory structure:
```sh
.
├── a.siva
├── b.siva
├── repo-a
└── some-dir
    ├── c.siva
    └── repo-b

```

`gitbase server -d=d1` -> will load `a.siva`,`b.siva`,`repo-a`,`repo-b`, `c.siva`
`gitbase server -d=d1 --no-git` -> will load `a.siva`, `b.siva`, `c.siva`
`gitbase server -d=d1 --no-siva`-> will load `repo-a`, `repo-b`
`gitbase server -d=d1 --depth=1` -> will load `a.siva`, `b.siva`, `repo-a`

All the variants combining `--no-git` and `--no-siva` flags with `--depth` are allowed and work too.
 
On the other hand, there is some useless code left in `repository_pool.go` about load directories, should I remove it?

@src-d/data-retrieval  WDYT?


repository_pool.go
```diff
@@ -262,6 +276,7 @@ func (p *RepositoryPool) addSivaFile(root, path string, f os.FileInfo) {
 
 	if strings.HasSuffix(f.Name(), ".siva") {
 		path := filepath.Join(path, f.Name())
+		fmt.Println(path)
```
erizocosmico: leftover?

repository_pool.go
```diff
@@ -182,6 +183,7 @@ func (p *RepositoryPool) AddGit(path string) (string, error) {
 // AddGitWithID checks if a git repository can be opened and adds it to the
 // pool. ID should be specified.
 func (p *RepositoryPool) AddGitWithID(id, path string) (string, error) {
+	fmt.Println(id, path)
```
erizocosmico: leftover?

cmd/gitbase/command/server.go
```diff
@@ -196,53 +200,119 @@ func (c *Server) registerDrivers() error {
...
+
+func isGitRepo(path string) (bool, error) {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return false, err
 	}
 
-	return nil
+	for _, file := range files {
+		if file.Name() == ".git" {
```
erizocosmico: what about bare repositories?

repository_pool.go
mcarmonaa: yes

repository_pool.go
mcarmonaa: yes

cmd/gitbase/command/server.go
mcarmonaa: fixed

eiso:
I love this! Great for user-friendliness.

mcarmonaa:
@ajnavarro sure, I'll do it too. Other than that, do you think I should remove all the code related to load repositories from `repository_pool.go`? Or do we want to keep it there to expose it although we are not using it anymore?

mcarmonaa:
- Documentation has been updated
- Dockerfile has been updated to use `--directories` flag
- Some code has been removed because is not used anymore

path_utils_test.go
```diff
@@ -12,124 +12,31 @@ func TestPatternMatches(t *testing.T) {
...
-			prefix, files, err := PatternMatches(test.path)
+			files, err := PatternMatches(test.path)
 			require.NoError(t, err)
-			require.Equal(t, test.prefix, prefix)
 			require.Exactly(t, test.expected, files)
 		})
 	}
 }
-
-func TestPatternPrefixDepth(t *testing.T) {
```
ajnavarro: why did you delete that test?

repository_pool.go
```diff
@@ -175,100 +174,33 @@ func (p *RepositoryPool) Add(repo repository) error {
...
-				}).Error("repository could not be added")
-			} else {
-				logrus.WithField("path", pa).Debug("repository added")
-			}
-		}
+// AddSivaFile adds to the pool the given file if it's a siva repository,
+// that is, has the .siva extension
+func (p *RepositoryPool) AddSivaFile(path string) error {
+	file := filepath.Base(path)
+	if !strings.HasSuffix(file, ".siva") {
```
ajnavarro: I would separate the logic related to check if the path is from a siva file, and add a siva repository to the repository pool. Per example, as we have a `isGitRepo` function, we should have a `isSivaFile` function to check if we can add that file to the pool.

path_utils_test.go
mcarmonaa: Because the function was tested here is not needed now so I removed it.

cmd/gitbase/command/server.go
```diff
@@ -196,53 +200,128 @@ func (c *Server) registerDrivers() error {
...
-	for _, m := range matches {
-		logrus.WithField("dir", m).Debug("siva repositories directory added")
-		if err := c.pool.AddSivaDir(m); err != nil {
+	initDepth := strings.Count(root, string(os.PathSeparator))
+	return filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
 			return err
 		}
+
+		if info.IsDir() {
```
smola: You could split these two cases to their own functions, and maybe then simplify some of the conditions with return early clauses.

mcarmonaa: I've placed the logic related to add a git repository in its own method, but I can see how to simplify the conditions

PR #419, opened 2018-08-10
Compare repos by ID.

kuba--:

The change forces decoder to compare repos by `ID` (not by `Path`). Thanks to this change `objectDecoder.decode` method doesn't have to create a new `RepoObjectDecoder` every time, but it re-uses internal cache.

It makes _full-scan_ queries for indexed columns much faster. For instance for 620 rows duration dropped from 14 sec. to 86 ms.

Closes: https://github.com/src-d/go-mysql-server/issues/324


kuba--:
@ajnavarro shall we merge it before release?

Issue #413, opened 2018-08-04
Can't process PGA due to nested folder structure


eiso:
Friction log for extracting all Python UASTs with Gitbase from a subset of PGA

```
pga list -l python -f json | head -n 10 | jq -r '.sivaFilenames[]' | pga get -i -o repositories

docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd

docker exec bblfshd bblfshctl driver install python docker://bblfsh/python-driver:v2.2.1

docker run -d --rm --name pilosa -p 10101:10101 pilosa/pilosa:v0.9.0

docker run -it --rm --name gitbase -p 3306:3306 --link pilosa:pilosa --link bblfshd -e PILOSA_ENDPOINT="http://pilosa:10101" -e BBLFSH_ENDPOINT="bblfshd:9432" -v $(pwd)/repositories/siva/latest:/opt/repos srcd/gitbase:latest
```

```
Errors
DEBU[0000] git repositories directory added              dir=/opt/repos
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/28: repository does not exist" id=28 path=/opt/repos/28
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/36: repository does not exist" id=36 path=/opt/repos/36
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/5b: repository does not exist" id=5b path=/opt/repos/5b
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/63: repository does not exist" id=63 path=/opt/repos/63
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/7c: repository does not exist" id=7c path=/opt/repos/7c
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/7f: repository does not exist" id=7f path=/opt/repos/7f
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/84: repository does not exist" id=84 path=/opt/repos/84
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/b7: repository does not exist" id=b7 path=/opt/repos/b7
ERRO[0000] repository could not be added                 error="the repository could not be opened: /opt/repos/ec: repository does not exist" id=ec path=/opt/repos/ec
[...]

PR #422, opened 2018-08-13
Update gitbase dependencies and turnoff proto logs.

kuba--:

This commit updates dependencies and registers a new index driver (`pilosalib`).
Because packages `go-pilosa` and `pilosa` share and register the same types in protobufs we have to double check if type was already registered. This is temporary fix to avoid logging _duplicate proto type registered_ when gitbase starts.


ajnavarro:
@kuba-- please, rebase with master, you have a commit from @erizocosmico

kuba--:
After merge into the master it closes: https://github.com/src-d/gitbase/issues/413

cmd/gitbase/command/server.go
```diff
@@ -193,6 +194,7 @@ func (c *Server) registerDrivers() error {
 	logrus.Debug("established connection with pilosa")
 
 	c.engine.Catalog.RegisterIndexDriver(pilosa.NewDriver(c.IndexDir, client))
+	c.engine.Catalog.RegisterIndexDriver(pilosalib.NewDriver(c.IndexDir))
```
ajnavarro: should we use another folder for that driver? maybe we can create a mess if we name two indexes in the same way from different drivers.

kuba--: We can do that, but in theory it should work, because it's a root folder where mapping and config file goes.  Mapping and Config are the same. The only concern I have if someone wants to create a 2 indexes for the same expression (but I think it's not possible).
So for old pilosa mapping and config goes to `root/db/table/id` with names `config.yml`, `mapping.db`. Config contains information about what driver was used to create an index.
An extra metadata for pilosalib goes under to `.pilosalib` folder under `id` folder.

So it's up to us. if we want to have 2 different roots per driver.

ajnavarro: I would prefer two folders to avoid unexpected problems.

kuba--: ok, fair enough

kuba--:
FYI: I modified `server.go`. Right now, if we cannot connect to pilosa server, we just log a warning (instead of returning an error) and skip registration (for this particular driver).
`pilosalib` driver will be registered by default (without any conditions).

PR #424, opened 2018-08-14
Add -g option for backward compatibility

kuba--:

We can handle `-g` option for backward compatibility with previous versions.
`-g` is not documented, but we can use it as a alias for `-d` option, e.g.:
`./gitbase server -g /tmp/boo -d /tmp/repos/ -i /tmp/indexes`.

It's done to let run regression tests with `regression-gitbase` (which by default uses `-g`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,6 @@ _cgo_export.*
 
 _testmain.go
 
-gitbase
 main
 *.exe
 *.test
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ _cgo_export.*
 
 _testmain.go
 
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -19,6 +19,7 @@ _cgo_export.*
 
 _testmain.go
 
+gitbase
```
erizocosmico: wont this exclude the gitbase folder in cmd?

smola: Yeah, also do not include gitignore changes in unrelated commits.

kuba--: Ok, fair enough - I've reverted gitignore.

PR #428, opened 2018-08-16
Regression queries

kuba--:

I'd like to add a file with regression queries. `regression-gitbase` will look into this file. If file doesn't exist (e.g. because of old version of gitbase) it will be tested against the default set of queries (like it worked before).

### DIFF
--- a/_testdata/regression.yml
+++ b/_testdata/regression.yml
@@ -7,12 +7,12 @@
   ID: 'query1'
   Name: 'Last commit messages in HEAD for every repository'
   Statements:
-    - SELECT c.commit_message FROM refs rJOIN commits c ON r.commit_hash = c.commit_hashWHEREr.ref_name = 'HEAD'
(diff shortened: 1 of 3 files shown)


jfontan:
I also believe that we should add some kind of unique ID. The format of it I'm not sure. If it's a string we can start with `Q1`, `Q2`... as @smola proposed and maybe move to something more meaningful in the future. Maybe grouping queries (`index1`, `uast7`).

kuba--:
Ok, I added IDs. In `regression-gitbase` it will be parsed to string, but right now, because I don't have a good proposal for naming convention (maybe it should start from semver), I've put just numbers.

erizocosmico:
Why do we need the version in the ID?

jfontan:
I believe that `Q-1`, `Q-2` would be good enough now. Does anyone has a better idea? @ajnavarro, @smola?

kuba--:
...renamed IDs to follow `queryX` convention.

kuba--:
@ajnavarro - added test to parse regression queries.

PR #430, opened 2018-08-17
Update go-git and go-mysql-server dependencies

ajnavarro:
Some refactoring was needed to be able to use the new packfile decoder API from go-git.

Speed performance on full scans and generating indexes of 40%-60%, depending on the dataset.


packfiles.go
```diff
@@ -193,7 +200,7 @@ func getUnpackedObject(repo repository, hash plumbing.Hash) (o object.Object, er
 type repoObjectDecoder struct {
 	repo     string
 	packfile plumbing.Hash
-	decoder  *packfile.Decoder
+	decoder  *packfile.Packfile
```
erizocosmico: should we rename `decoder` to `packfile` then?

kuba--: ... and remove/rename `packfile plumbing.Hash`?

vendor/github.com/pilosa/go-pilosa/gopilosa_pbuf/public.pb.go
```diff
@@ -544,42 +544,18 @@ func (m *ImportValueRequest) GetValues() []int64 {
...
-		proto.RegisterType((*Pair)(nil), "internal.Pair")
-	}
-	if t := proto.MessageType("internal.ValCount"); t == nil {
-		proto.RegisterType((*ValCount)(nil), "internal.ValCount")
-	}
-	if t := proto.MessageType("internal.Bit"); t == nil {
-		proto.RegisterType((*Bit)(nil), "internal.Bit")
-	}
-	if t := proto.MessageType("internal.ColumnAttrSet"); t == nil {
-		proto.RegisterType((*ColumnAttrSet)(nil), "internal.ColumnAttrSet")
```
kuba--: good to keep these extra ifs. it avoids printing warning about already registered type, on start.

erizocosmico: Vendor files should not be modified manually

kuba--: We'll have to find a better way for these autogenerated files, but so far I modified them adding extra ifs to avoid nasty warnings (after integrating pilosalib driver - both try to register the same types in grpc)

ajnavarro: Actually, I didn't have protobuf errors on startup anymore. Some dependency must be updated.

kuba--: @ajnavarro - do you have pilosa running? Right now, if pilosa is not running we don't register old driver. The warnings appeared when we registered both drivers, because `pilosa` and `go-pilosa` register in protobuf the same types.

@erizocosmico - I totally agree, but it was just ad-hoc _fix_ to suppress ugly warnings. It was temporary, the quickest fix.

Moreover based on comments in protobufs:
https://github.com/golang/protobuf/blob/master/proto/properties.go#L481
Some day _duplicate proto type register_ may panic.

ajnavarro:
Hmm, I think CI is failing because now the objects are iterated in another order. Can that be the cause @erizocosmico @jfontan ? WDYT?

erizocosmico:
@ajnavarro it can be that, the errors in travis are not very descriptive, though, so no way to be certain.

jfontan:
The problem was the order of the objects returned by the iterator. Sent a PR changing the iterator back to offset order.

src-d/go-git#927

PR #431, opened 2018-08-22
parse and analyze regression queries

kuba--:

### DIFF
--- a/_testdata/regression.yml
+++ b/_testdata/regression.yml
@@ -29,11 +29,13 @@
   Name: 'Create pilosa index on language UDF'
   Statements:
     - CREATE INDEX language_idx ON files USING pilosa (language(file_path, blob_content)) WITH (async = false)
+    - DROP INDEX language_idx ON files
 -
   ID: 'query6'
   Name: 'Create pilosalib index on language UDF'
   Statements:
     - CREATE INDEX language_idx ON files USING pilosalib (language(file_path, blob_content)) WITH (async = false)
+    - DROP INDEX language_idx ON files
 -
   ID: 'query7'
   Name: 'Query by language using the pilosa index'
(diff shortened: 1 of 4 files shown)


ajnavarro:
Why this query didn't fail before @kuba-- ?

kuba--:
@ajnavarro it started failing ~ 5 days ago (since 1b182d7a429175d75de8c4a333c4a24a168fe1ca commit to regression-gitbase). The last time when it succeeded was 8 days ago. Around this time we released 16-RC1.

ajnavarro:
But we added some smoke tests on gitbase as well, right? That tests are intended to tell us when we break that queries and avoid that.

kuba--:
What I've added was just syntax checker (to check that all statements in yaml file are correct SQL queries). I don't evaluate them.
But looks that we also have to add evaluators for all these queries.

kuba--:
I've just added `Analyzer`. So, we can parse and analyze statements. Right now, problems like _renamed column name_ will be caught.

ajnavarro:
What do you think about call directly to the engine, as we are doing on `integration_test.go` file? I think it will be more error prune. WDYT? (We should have only one place where we are initializing the engine BTW, right now we have two places; `integration_test.go` and `server.go`)

kuba--:
Right now I'm doing exactly the same as it's done in `engine.Query`.
The only difference is that I don't call:
```
	iter, err := analyzed.RowIter(ctx)
	analyzed.Schema()
```

- I can rewrite `regresssion_test` in the way how `integration_test` is done (but it won't be a smoke test but something what will reuse `fixtures "gopkg.in/src-d/go-git-fixtures.v3"`.
- Regarding server instances. Maybe it's doable to initialize engine only in the server and in any other places (like tests) we can create a Server instance. Does it make sense?

ajnavarro:
1 - you're right, it will execute queries on top of a real repository, making that test an integration test more than a smoke test, but I think is worth it to avoid future incompatibilities or errors on gitbase-regression.
2 - Makes a lot of sense.

kuba--:
Now regression queries are executed.

Issue #436, opened 2018-08-22
Optimal layout


kuba--:
We can save a few bytes with optimal structure padding, e.g.:
```
squash_iterator.go:43:22: struct of size 80 could be 72
{
row,
filters,
ctx,
iter,
repo,
done,
skipGitErrors
}

squash_iterator.go:409:20: struct of size 112 could be 104
{
row
filters
refs
ctx
repo
head
ref
repos
virtual
skipGitErrors
}

squash_iterator.go:3130:28: struct of size 120 could be 112
{
row,
commits
filters
files
file
commit
ctx
treeHash
skipGitErrors
}

squash_iterator.go:3257:33: struct of size 128 could be 120
{
row
filters
index
repo
iter
file
pool
ctx
treeHash
skipGitErrors
}
```

WDYT?

PR #437, opened 2018-08-24
Reorganise structs layout to improve mem padding

kuba--:
Closes https://github.com/src-d/gitbase/issues/436

This is a small thing, but maybe worth to consider (WDYT)?. As an example I re-organized couple structs to make memory alignment optimal - less work for GC.

Changes:
- `Server` size was 192, now 176 
- `squashReposIter` size was 80, now 72
- `squashRefIter` size was size 112, now 104
- `squashCommitFilesIter` size was 120, now 112
- `squashIndexCommitFilesIter` size was 128, now 120

### DIFF
--- a/cmd/gitbase/command/server.go
+++ b/cmd/gitbase/command/server.go
@@ -35,11 +35,13 @@ const (
 
 // Server represents the `server` command of gitbase cli tool.
 type Server struct {
-	Verbose       bool     `short:"v" description:"Activates the verbose mode"`
+	engine *sqle.Engine
+	pool   *gitbase.RepositoryPool
+	name   string
+
+	Version       string   // Version of the application.
 	Directories   []string `short:"d" long:"directories" description:"Path where the git repositories are located (standard and siva), multiple directories can be defined. Accepts globs."`
 	Depth         int      `long:"depth" default:"1000" description:"load repositories looking at less than <depth> nested subdirectories."`
-	DisableGit    bool     `long:"no-git" description:"disable the load of git standard repositories."`
-	DisableSiva   bool     `long:"no-siva" description:"disable the load of siva files."`
 	Host          string   `long:"host" default:"localhost" description:"Host where the server is going to listen"`
 	Port          int      `short:"p" long:"port" default:"3306" description:"Port where the server is going to listen"`
 	User          string   `short:"u" long:"user" default:"root" description:"User name used for connection"`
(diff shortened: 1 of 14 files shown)


kuba--:
@mcarmonaa - couple readings:
- https://dave.cheney.net/2015/10/09/padding-is-hard
- https://go101.org/article/memory-layout.html
- http://www.catb.org/esr/structure-packing/
- https://syslog.ravelin.com/go-and-memory-layout-6ef30c730d51

(or just by searching about go memory alignment), but briefly - memory aligns to 8 bytes (most likely but it depends on architecture), so if you layout your structure like:
```
type MyData struct {
	a bool
	b int64
	c int32
	d bool
	e string
}
```
because of 8bytes alignment, compiler will introduce memory padding, so it will look:
```
MyData.a bool: 0-1 (size 1, align 1)
padding: 1-8 (size 7, align 0) # <---------------------
MyData.b int64: 8-16 (size 8, align 8)
MyData.c int32: 16-20 (size 4, align 4)
MyData.d bool: 20-21 (size 1, align 1)
padding: 21-24 (size 3, align 0) # <--------------------
MyData.e string: 24-40 (size 16, align 8)
```
and thats why you'll loose 10 bytes.

But if you layout your structure for instance as follows:
```
type MyData struct {
	e string
	b int64
	c int32
	a bool
	d bool
}
```
you'll loose just 2 bytes (I think the last padding is OK):
```
MyData.e string: 0-16 (size 16, align 8)
MyData.b int64: 16-24 (size 8, align 8)
MyData.c int32: 24-28 (size 4, align 4)
MyData.a bool: 28-29 (size 1, align 1)
MyData.d bool: 29-30 (size 1, align 1)
padding: 30-32 (size 2, align 0) # <----------
```

Generally, the simplest thing what I do I just layout structure from the biggest type to the smallest.

mcarmonaa:
Thank you very much @kuba-- for such a good explanation and the links! :sparkles:

cmd/gitbase/command/server.go
```diff
@@ -49,14 +51,11 @@ type Server struct {
...
 	ReadOnly      bool     `short:"r" long:"readonly" description:"Only allow read queries. This disables creating and deleting indexes as well." env:"GITBASE_READONLY"`
-	// SkipGitErrors disables failing when Git errors are found.
-	SkipGitErrors bool
-	// Version of the application.
-	Version string
 
-	engine *sqle.Engine
-	pool   *gitbase.RepositoryPool
-	name   string
+	SkipGitErrors bool // SkipGitErrors disables failing when Git errors are found.
```
ajnavarro: This is changing the output of `gitbase server --help` command? If yes, could you change it on the documentation too please?

kuba--: @ajnavarro - done

PR #439, opened 2018-08-27
Improvement/upgrade bblfshd client

jfontan:
Upgrade bblfsh client and add support to select mode when using `uast`.

```
uast(blob, "python", "//Identifier", "semantic")
```

### DIFF
--- a/internal/function/uast.go
+++ b/internal/function/uast.go
@@ -8,7 +8,6 @@ import (
 	"github.com/src-d/gitbase"
 	bblfsh "gopkg.in/bblfsh/client-go.v2"
 	"gopkg.in/bblfsh/client-go.v2/tools"
-	"gopkg.in/bblfsh/sdk.v1/protocol"
 	"gopkg.in/bblfsh/sdk.v1/uast"
 
 	errors "gopkg.in/src-d/go-errors.v1"
@@ -26,12 +25,11 @@ type UAST struct {
 	Blob  sql.Expression
 	Lang  sql.Expression
 	XPath sql.Expression
-	Mode  sql.Expression
 }
 
 // NewUAST creates a new UAST UDF.
 func NewUAST(args ...sql.Expression) (sql.Expression, error) {
-	var blob, lang, xpath, mode sql.Expression
+	var blob, lang, xpath sql.Expression
 	switch len(args) {
 	case 1:
 		blob = args[0]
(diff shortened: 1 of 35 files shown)


jfontan:
@bzz updated bblfsh minimum version in the documentation

internal/function/uast_test.go
```diff
@@ -97,6 +113,7 @@ func TestUASTXPath(t *testing.T) {
 		{"left is nil", sql.NewRow(nil, "foo"), nil, nil},
 		{"right is nil", sql.NewRow(uast, nil), nil, nil},
 		{"both given", sql.NewRow(uast, testXPath), filteredNodes, nil},
+		// {"blob is invalid", fn4, sql.NewRow([]byte(testCode), "Python", testXPath, "invalid"), nil},
```
mcarmonaa: leftover?

internal/function/uast.go
```diff
@@ -184,6 +202,39 @@ func (f UAST) Eval(ctx *sql.Context, row sql.Row) (out interface{}, err error) {
 		xpath = x.(string)
 	}
 
+	modeSet := false
```
ajnavarro: maybe is worth it to move this to a private function to improve readability?

erizocosmico:
What happens if you want semantic uast but not want to pass language or xpath query? Being the last parameter forces you to give the other two.

ajnavarro:
@erizocosmico you're right. Maybe we will need another UDF for semantic UAST (maybe `SUAST`? :D)

internal/function/uast.go
```diff
@@ -40,10 +42,15 @@ func NewUAST(args ...sql.Expression) (sql.Expression, error) {
 		blob = args[0]
 		lang = args[1]
 		xpath = args[2]
+	case 4:
```
kuba--: I know it's a silly detail, but how about:
```
switch len(args) {
       default: return nil, sql.ErrInvalidArgumentNumber.New("1, 2, 3 or 4", len(args))
 	case 4:		 	
 		mode = args[3]
                fallthrough
 	case 3:		 	
 		xpath = args[2]
                fallthrough
 	case 2:		 	 		
 		lang = args[1]	
                fallthrough
 	case 1:
		blob = args[0]
}
```
less copy paste - up to you ;)

internal/function/uast.go
jfontan: Is not used anymore as it is not an optional parameter.

internal/function/uast.go
jfontan: I've left it was it was previously as `uast` now changed back to its original parameters. We can do a second round of refinement if we feel it's needed.

jfontan:
Implemented `uast_mode(blob, language, mode)` and `uast` now has the same parameters as before but returns `semantic` version of AST.

jfontan:
The last change moves the `mode` of `uast_mode` to the first parameter. If we want to make `lang` optional in the future we can do it without breaking compatibility.

PR #453, opened 2018-08-31
Feature/new uast serialization

mcarmonaa:
Instead of serialize UASTs as a JSON array of marshaled UAST nodes encoded as base64, now the list of nodes are serialized as a sequence of marshaled nodes preceded by their lengths, it is:
```
BigEndianInt32(len(marhsal(node))+marshal(node)+
BigEndianInt32(len(marhsal(node))+marshal(node)+
BigEndianInt32(len(marhsal(node))+marshal(node)+
...
```
The node size is being encoded as an int32 in big endian to make it directly readable from the JVM just in case we will need manage it from spark. If you think there are better options just let me know and I'll change it.

Also, there is a new flag in the cli `--old-uast-serialization` and an environment variable `GITBASE_UAST_SERIALIZATION` to choose if this mode or the old one. This way, we don't brake compatibility for the moment.

### DIFF
--- a/cmd/gitbase/command/server.go
+++ b/cmd/gitbase/command/server.go
@@ -56,6 +56,7 @@ type Server struct {
 	DisableGit    bool `long:"no-git" description:"disable the load of git standard repositories."`
 	DisableSiva   bool `long:"no-siva" description:"disable the load of siva files."`
 	Verbose       bool `short:"v" description:"Activates the verbose mode"`
+	OldUast       bool `long:"old-uast-serialization" description:"serialize uast in the old format" env:"GITBASE_UAST_SERIALIZATION"`
 }
 
 type jaegerLogrus struct {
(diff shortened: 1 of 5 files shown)


erizocosmico:
I think given this is a v0 we could actually break compatibility and just enable this mode. WDYT @ajnavarro?

ajnavarro:
@erizocosmico yep, the idea is to use this new format as default, but provide a way to make it work with dependant applications providing the flag (the flag will be removed in next versions).

cmd/gitbase/command/server.go
```diff
@@ -56,6 +56,7 @@ type Server struct {
 	DisableGit    bool `long:"no-git" description:"disable the load of git standard repositories."`
 	DisableSiva   bool `long:"no-siva" description:"disable the load of siva files."`
 	Verbose       bool `short:"v" description:"Activates the verbose mode"`
+	AltUast       bool `long:"alt-uast-serialization" description:"serialize uast in a new format" env:"GITBASE_UAST_SERIALIZATION"`
```
ajnavarro: could you change the flag to do the opposite? it should active old serialization format.

ajnavarro:
Pinging @bzz @carlosms to make them aware of that change (anyways, on new release notes we will add a paragraph explaining this breaking change).

bzz:
Thank you for kind notice @ajnavarro!  @carlosms thank you for taking care of it

PR #457, opened 2018-09-05
internal/function: use cache for expensive language operations 

### LINKED ISSUE #442: Add a intermediate cache for language UDF
language UDF can be very slow when querying a lot of files. Sometimes this is executing language for the same file multiple times. For example, if we iterate the `files` table the same file is multiplied by the number of commits.

It could be good to add an intermediate cache so the results from previous run can be used instead of finding again the language. We can use an LRU using the file name and blob hashes. We may be able to use the hash of the first bytes of the blob that enry uses to speed up the process.

erizocosmico:
Closes #442 

This adds a two queue lru cache to language UDF. Why a 2Q cache instead of a regular LRU? Because we do not only want the most recent but the most frequent as well to maximize the cache usage.

Since running enry on a file_path only is just marginally slower than using the cache, I decided to not use the cache unless the blob_content is being provided (we could use two caches, though) to not pollute the cache and use it only for the really expensive operations: reading stuff from blobs.

Note that the cache is per gitbase instance.

Results are pretty much the ones @ajnavarro describes in the issue.

Before:

```
mysql> select count(*) from (select file_path, language(file_path, blob_content) as language FROM files WHERE language(file_path, blob_content) = 'Go') t;
+----------+
| COUNT(*) |
+----------+
|    39781 |
+----------+
1 row in set (20.85 sec)
```

Now:

```
mysql> select count(*) from (select file_path, language(file_path, blob_content) as language FROM files WHERE language(file_path, blob_content) = 'Go') t;
+----------+
| COUNT(*) |
+----------+
|    39781 |
+----------+
1 row in set (17.14 sec)
```

### DIFF
--- a/docs/using-gitbase/configuration.md
+++ b/docs/using-gitbase/configuration.md
@@ -12,6 +12,7 @@
 | `GITBASE_INDEX_DIR`          | directory to save indexes, default `/var/lib/gitbase/index`                        |
 | `GITBASE_TRACE`              | enable jaeger tracing, default disabled                                            |
 | `GITBASE_READONLY`           | allow read queries only, disabling creating and deleting indexes, default disabled |
+| `GITBASE_LANGUAGE_CACHE_SIZE`           | size of the cache for the `language` UDF. The size is the maximum number of elements kept in the cache, 10000 by default |
 
 ### Jaeger tracing variables
 
--- a/internal/function/language.go
+++ b/internal/function/language.go
@@ -1,11 +1,26 @@
 package function
 
 import (
+	"encoding/binary"
 	"fmt"
+	"hash/crc32"
 
+	lru "github.com/hashicorp/golang-lru"
 	enry "gopkg.in/src-d/enry.v1"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
+
+const defaultLanguageCacheSize = 10000
+
+var languageCache *lru.TwoQueueCache
+
+func init() {
+	var err error
+	languageCache, err = lru.New2Q(defaultLanguageCacheSize)
+	if err != nil {
+		panic(fmt.Errorf("cannot initialize language cache: %s", err))
+	}
+}
 
 // Language gets the language of a file given its path and
 // the optional content of the file.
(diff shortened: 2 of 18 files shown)


internal/function/language.go
```diff
@@ -1,12 +1,27 @@
...
+	"encoding/binary"
 	"fmt"
+	"hash/crc32"
 
+	lru "github.com/hashicorp/golang-lru"
 	enry "gopkg.in/src-d/enry.v1"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
 
+const defaultLanguageCacheSize = 10000
```
kuba--: Right now, we are trying to let apps adjust cache sizes (and pass them down to packages like go-git), so maybe in this case we can also think about how easily we can modify this cache in the future (at least size).

erizocosmico: Yep, this should be configurable, right @ajnavarro? Environment variable or CLI flag that passes down to here? WDYT?

ajnavarro: yep, it should be configurable. it should be a env var o cli flag, yep.

erizocosmico:
Added configuration to the cache size via env var

docs/using-gitbase/configuration.md
```diff
@@ -12,6 +12,7 @@
 | `GITBASE_INDEX_DIR`          | directory to save indexes, default `/var/lib/gitbase/index`                        |
 | `GITBASE_TRACE`              | enable jaeger tracing, default disabled                                            |
 | `GITBASE_READONLY`           | allow read queries only, disabling creating and deleting indexes, default disabled |
+| `GITBASE_LANGUAGE_CACHE_SIZE`           | size of the cache for the `language` UDF. The size is the maximum number of elements kept in the cache |
```
ajnavarro: shall we add the default value here?

internal/function/language.go
```diff
@@ -1,12 +1,42 @@
...
+	"os"
+	"strconv"
 
+	lru "github.com/hashicorp/golang-lru"
 	enry "gopkg.in/src-d/enry.v1"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
 
+const (
+	languageCacheSizeKey     = "GITBASE_LANGUAGE_CACHE_SIZE"
```
ajnavarro: should we add that env variable as we are doing with others? check https://github.com/src-d/gitbase/blob/master/cmd/gitbase/command/server.go#L50

erizocosmico: If we add it there we will have to pass it all the way down until the infinite. Besides, there are other examples of environment variables such as blob max size that are not there but where they're used

erizocosmico:
Added default value to docs

PR #461, opened 2018-09-06
internal/function: add LRU cache for uast and uast_mode functions

mcarmonaa:
This PR adds ~~two caches~~ one cache for the functions `uast` and  `uast_mode`. 

The main cache is set at server level, so it is shared by all the queries. It's caching UASTs requested to bblfsh

~~Meanwhile, the second cache is used at query level and it caches the parsed errors from bblfsh. I thought it makes sense in the context of the same query, that if bblfsh returns a parsed error for a blob, most probably it will return the same error for that blob during the execution of the that query.~~

Results of theses changes running this query
> select file_path,uast(blob_content,language(file_path)) from files limit 100000;

on  a set of 20 siva-files with cache sizes of `10000` are:
- `12 min 43.18 sec` -> NO CACHE
- `4 min 13.12 sec` -> CACHE
- ~~`12 min 30.88 sec` -> NO CACHE~~
- ~~`3 min 56.06 sec` -> CACHE UASTs~~
- ~~`1 min 2.60 sec` -> CACHE UASTs and Errors~~
- ~~`59.21 sec` -> CACHE UASTs and Errors with new uast serialization~~

Other non-directly related change is that I've refactored the code for `UAST` and `UASTMode` functions. They are almos the same, so the code was starting to be too much duplicated. Now both of them have an `uastFunc` struct embedded which keeps almost all the implementation. The `UAST` and `UASTMode` functions just implement the constructors and the `TransformUp` and `String` methods.

### DIFF
--- a/internal/function/uast.go
+++ b/internal/function/uast.go
@@ -4,6 +4,8 @@ import (
 	"crypto/sha1"
 	"fmt"
 	"hash"
+	"os"
+	"strconv"
 	"strings"
 
 	lru "github.com/hashicorp/golang-lru"
@@ -16,13 +18,21 @@ import (
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 )
 
-const defaultUASTCacheSize = 10000
+const (
+	uastCacheSize        = "GITBASE_UAST_CACHE_SIZE"
+	defaultUASTCacheSize = 10000
+)
 
 var uastCache *lru.Cache
 
 func init() {
-	var err error
-	uastCache, err = lru.New(defaultUASTCacheSize)
+	s := os.Getenv(uastCacheSize)
+	size, err := strconv.Atoi(s)
+	if err != nil || size <= 0 {
+		size = defaultUASTCacheSize
+	}
+
+	uastCache, err = lru.New(size)
 	if err != nil {
 		panic(fmt.Errorf("cannot initialize UAST cache: %s", err))
 	}
@@ -246,7 +256,7 @@ func NewUAST(args ...sql.Expression) (sql.Expression, error) {
 
 	switch len(args) {
 	default:
-		return nil, sql.ErrInvalidArgumentNumber.New("1, 2, or 3", len(args))
+		return nil, sql.ErrInvalidArgumentNumber.New("1, 2 or 3", len(args))
 	case 3:
 		xpath = args[2]
 		fallthrough
--- a/internal/function/uast.go
+++ b/internal/function/uast.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	lru "github.com/hashicorp/golang-lru"
-	"github.com/sirupsen/logrus"
 	bblfsh "gopkg.in/bblfsh/client-go.v2"
 	"gopkg.in/bblfsh/client-go.v2/tools"
 	"gopkg.in/bblfsh/sdk.v1/uast"
@@ -53,8 +52,7 @@ type uastFunc struct {
 	Lang  sql.Expression
 	XPath sql.Expression
 
-	h        hash.Hash
-	errCache *lru.Cache
+	h hash.Hash
 }
 
 // IsNullable implements the Expression interface.
@@ -207,18 +205,10 @@ func (u *uastFunc) getUAST(
 	if ok {
 		node = value.(*uast.Node)
 	} else {
-		if u.errCache != nil {
-			_, ok := u.errCache.Get(key)
-			if ok {
-				return nil, nil
-			}
-		}
-
 		var err error
 		node, err = getUASTFromBblfsh(ctx, blob, lang, xpath, mode)
 		if err != nil {
 			if ErrParseBlob.Is(err) {
-				u.errCache.Add(key, struct{}{})
 				return nil, nil
 			}
 
@@ -267,18 +257,12 @@ func NewUAST(args ...sql.Expression) (sql.Expression, error) {
 		blob = args[0]
 	}
 
-	errCache, err := lru.New(defaultUASTCacheSize)
-	if err != nil {
-		logrus.Warn("couldn't initialize UAST cache for errors")
-	}
-
 	return &UAST{&uastFunc{
-		Mode:     mode,
-		Blob:     blob,
-		Lang:     lang,
-		XPath:    xpath,
-		h:        sha1.New(),
-		errCache: errCache,
+		Mode:  mode,
+		Blob:  blob,
+		Lang:  lang,
+		XPath: xpath,
+		h:     sha1.New(),
 	}}, nil
 }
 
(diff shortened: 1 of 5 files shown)


internal/function/uast.go
```diff
@@ -1,9 +1,13 @@
 package function
 
 import (
+	"crypto/sha1"
 	"fmt"
+	"hash"
 	"strings"
 
+	lru "github.com/hashicorp/golang-lru"
```
kuba--: I thought about: `github.com/golang/groupcache/lru`, but if hashicorp works for you that's ok.

internal/function/uast.go
```diff
@@ -139,141 +162,176 @@ func (f UAST) Eval(ctx *sql.Context, row sql.Row) (out interface{}, err error) {
...
+	*uastFunc
+}
+
+var _ sql.Expression = (*UAST)(nil)
+
+// NewUAST creates a new UAST UDF.
+func NewUAST(args ...sql.Expression) (sql.Expression, error) {
+	var mode = expression.NewLiteral("semantic", sql.Text)
+	var blob, lang, xpath sql.Expression
+	switch len(args) {
```
kuba--: how about (less copy-paste):
```go
switch len(args) {
default:
 		return nil, sql.ErrInvalidArgumentNumber.New("1, 2, or 3", len(args))
case 3: 
                xpath = args[2]
                fallthrough
case 2:
               lang = args[1]
               fallthrough
case 1:
               blob = args[0]
}
```

internal/function/uast.go
mcarmonaa: It seems that groupcache/lru is not thread safe.

> Cache is an LRU cache. It is not safe for concurrent access.

[groupcache/lru godoc](https://godoc.org/github.com/golang/groupcache/lru)

internal/function/uast.go
```diff
@@ -139,141 +162,176 @@ func (f UAST) Eval(ctx *sql.Context, row sql.Row) (out interface{}, err error) {
...
+		}
 
-// Resolved implements the Expression interface.
-func (f UASTMode) Resolved() bool {
-	return f.Blob.Resolved() && f.Lang.Resolved() && f.Mode.Resolved()
-}
+		var err error
+		node, err = getUASTFromBblfsh(ctx, blob, lang, xpath, mode)
+		if err != nil {
+			if ErrParseBlob.Is(err) {
```
ajnavarro: I think this is wrong. You are caching any kind of error, even connection ones.

internal/function/uast.go
```diff
@@ -12,109 +16,111 @@ import (
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 )
 
+const defaultUASTCacheSize = 10000
```
ajnavarro: this should be configurable by an env var or cli flag.

mcarmonaa:
related to @ajnavarro 's https://github.com/src-d/gitbase/pull/461#discussion_r215614022

***SOME NOTES ABOUT CACHING ERRORS***

I've been trying to classify the errors coming form bblfsh to know if they should be cached or not and I've found quite difficult since all of them are just a grpc error and an attached string.

Having as a reference measure the time `100000 rows in set (4 min 31.78 sec)` obtained without errors cache, if the most generic errors I found are cached
```go
strings.Contains(err.Error(), "could not autodetect language") ||
strings.Contains(err.Error(), "cannot decode uast") ||
strings.Contains(err.Error(), "missing driver for language") 
```
the speed decrease to `100000 rows in set (4 min 57.09 sec)`

Adding the error:
```
strings.Contains(err.Error(), "context deadline exceeded")
```
to the cached errors it improves to `100000 rows in set (2 min 12.11 sec)`, but I'm not sure if I should assume that during the execution of a query a blob returning this error will return it again.

One "funny" thing I realised about the `context deadline exceeded` is that I always get this error coming from parsing javascript files.

On the other hand, there are errors that show a trace I guess returned by the native driver and, as they are right now, are totally unclassifiable. They look like:
```
WARN[0036] unable to parse the given blob using bblfsh: rpc error: code = Unknown desc = unexpected token tCOLON


WARN[0026] unable to parse the given blob using bblfsh: rpc error: code = Unknown desc = undefined method `loc' for 1:Integer


WARN[0025] unable to parse the given blob using bblfsh: rpc error: code = Unknown desc = unexpected token $end

WARN[0048] unable to parse the given blob using bblfsh: rpc error: code = Unknown desc = Traceback (most recent call last):
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/requestprocessor.py", line 151, in process_request
    ast = AstImprover(code, orig_ast).parse()
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 38, in parse
    res = self.visit(self._astdict, root=True)
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 65, in visit
    visit_result = meth(node)
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 248, in visit_other
    node[field] = meth(child)
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 255, in visit_other_field
    return [self.visit(x) for x in node]
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 255, in <listcomp>
    return [self.visit(x) for x in node]
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 65, in visit
    visit_result = meth(node)
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 248, in visit_other
    node[field] = meth(child)
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 253, in visit_other_field
    return self.visit(node)
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 65, in visit
    visit_result = meth(node)
  File "/opt/driver/bin/.local/lib/python3.6/site-packages/python_driver/astimprove.py", line 221, in visit_arguments
    kwarg["ast_type"] = "kwarg"
TypeError: 'str' object does not support item assignment

```

If they were tagged like `unable to parse the given blob using bblfsh: rpc error: code = Unknown desc = NATIVE PARSE ERROR: whatever trace`, I'd bet that we could improve the speed.

mcarmonaa:
Another thing about speedup the operations on the cache, we could pass to the functions `uast` and `uast_mode` other argument containing the `blob_hash` so it wouldn't be necessary calculate it to get the key for the cache.

We could use directly the string concatenation of the `blob_hash`+`mode`+`language`.

internal/function/uast_utils.go
```diff
@@ -192,6 +214,10 @@ func nodesFromBlobArray(data interface{}) ([]*uast.Node, error) {
 	}
 
 	arr := data.([]interface{})
+	if len(arr) == 0 {
```
ajnavarro: this check is needed?

internal/function/uast_utils.go
```diff
@@ -219,6 +245,10 @@ func nodesFromBlob(data interface{}) ([]*uast.Node, error) {
 }
 
 func unmarshalNodes(data []byte) ([]*uast.Node, error) {
+	if len(data) == 0 {
```
ajnavarro: the same as previous comment

internal/function/uast_utils.go
mcarmonaa: Tests for the old serialization checked that uast and uast_mode returned nil in some cases. Without these checks those cases return a slice of length 0 for the new serialization.

internal/function/uast_utils.go
mcarmonaa: answered in the previous comment

PR #462, opened 2018-09-06
update docker installation text

tsolakoua:
I hadn't noticed that ```/my/git/repos``` should be replaced by my git directory until @jfontan told me, and I thought it might be useful for users that might not read carefully the command.

Signed-off-by: tsolakou <tsolakou.a@gmail.com>

### DIFF
--- a/docs/using-gitbase/getting-started.md
+++ b/docs/using-gitbase/getting-started.md
@@ -39,7 +39,7 @@ $ make dependencies
 
 ## Running with docker
 
-You can use the official image from [docker hub](https://hub.docker.com/r/srcd/gitbase/tags/) to quickly run gitbase:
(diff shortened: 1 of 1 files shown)


erizocosmico:
There's a note at the end of the docker section https://github.com/src-d/gitbase/blob/master/docs/using-gitbase/getting-started.md#running-with-docker, but maybe we can also add it there as well.

tsolakoua:
I hadn't noticed that note until now, since I didn't go to pilosa at all and I missed it. But I think it looks better as it's already is with the note at the end. If we add it where I added in the PR as well, then we will have to add it to the pilosa to keep consistency and it's too much.

erizocosmico:
@tsolakoua don't worry, the PR that added that was just merged, so it wasn't there before.

Ok, we can close the issue then

tsolakoua:
Thanks @erizocosmico

Issue #440, opened 2018-08-27
Increase go-git cache size


jfontan:
Right now there is no way of changing the default cache size in `go-git` and its size too small (96 MiB). I've been doing tests changing this value and its performance improved a lot.

Repositories: linux (2013), numpy, tensorflow
Number of rows: 395709
Query: `SELECT count(*) FROM commits c NATURAL JOIN ref_commits r WHERE r.ref_name = 'HEAD';`

Default cache: `1 row in set (54 min 22.17 sec)`
Cache size * 8: `1 row in set (20 min 43.69 sec)`

Memory consumption is also not too big. `gitbase` used 1.3 GiB in this query.

We should add an option to go-git Open to select cache size.

eiso:
@jfontan how does this scale with increased memory in relation to repo size? The reason I am asking is that I much rather have a significantly faster response and have a customer use 100GB of ram for a query. High memory instances are cheap these days.

jfontan:
@eiso, maybe it could heuristically assign a size but the fist step will be adding an option to change it.

kuba--:
...or maybe we can start from something like 20% of RAM as default for gitbase

kuba--:
Maybe we should be able to configure cache size not only for `PlainOpen`, but everywhere where we use decoders/encoders (`packfiles.go`). For instance when we're creating a new _repo decoder_, we call:
`packfile := packfile.NewPackfile(idx, fs, packf)` what under the hood creates a default object LRU cache.  Maybe we can create there NewPackfileWithCache and pass the same cache option from Storage - WDYT?

eiso:
@jfontan my question was more about if the memory improvement stays the same dependent on size of the repository (which I am assuming is a yes)

kuba--:
I'll try to implement it by adding support for couple environment variables, e.g.:
`GO_GIT_OBJECT_MAXCACHE` etc.
So everybody will be able to adjust go-git cache, doesn't matter if it's used directly or via apps like gitbase.
Any thoughts/concerns?

I think it's more flexible (can skip couple levels of passing options from one constructor to another) and it doesn't require any extra integration.
For instance you can just upgrade go-git in gitbase without any other changes in implementation and by setting env vars, benefit from bigger cache.

jfontan:
@eiso unfortunately I've seen changes in speedup from repo to repo but I don't know the cause. Maybe is the repo that has more complex commit tree and makes iterating it more expensive. We would have to measure more to know for sure.

[...]

PR #464, opened 2018-09-10
One configurable cache per `repository pool`

kuba--:
This PR closes https://github.com/src-d/gitbase/issues/440

Changes:
- Updated `go-git`
- `repository` interface requires `Cache() cache.Object`
- one cache object per `RepositoryPool`


cmd/gitbase/command/server.go
```diff
@@ -39,18 +40,19 @@ type Server struct {
...
+	Depth         int            `long:"depth" default:"1000" description:"load repositories looking at less than <depth> nested subdirectories."`
+	Host          string         `long:"host" default:"localhost" description:"Host where the server is going to listen."`
+	Port          int            `short:"p" long:"port" default:"3306" description:"Port where the server is going to listen."`
+	User          string         `short:"u" long:"user" default:"root" description:"User name used for connection."`
+	Password      string         `short:"P" long:"password" default:"" description:"Password used for connection."`
+	PilosaURL     string         `long:"pilosa" default:"http://localhost:10101" description:"URL to your pilosa server." env:"PILOSA_ENDPOINT"`
+	IndexDir      string         `short:"i" long:"index" default:"/var/lib/gitbase/index" description:"Directory where the gitbase indexes information will be persisted." env:"GITBASE_INDEX_DIR"`
+	CacheSize     cache.FileSize `long:"cache" default:"536870912" description:"Object cache size" env:"GITBASE_CACHE_SIZE"`
```
ajnavarro: could you change the default to something more human-friendly? like 100 and allows only MB

kuba--: yeah sure, how about 512MB?

jfontan: Also change the units of the parameter to MiB, That is, to specify 100 MiB you set `--cache 100`.

repository_pool.go
```diff
@@ -144,17 +148,23 @@ func (r *sivaRepository) Path() string {
...
+	cache        cache.Object
 }
 
-// NewRepositoryPool initializes a new RepositoryPool
-func NewRepositoryPool() *RepositoryPool {
+// NewRepositoryPool initializes a new RepositoryPool with LRU cache.
+func NewRepositoryPool(maxCacheSize cache.FileSize) *RepositoryPool {
 	return &RepositoryPool{
 		repositories: make(map[string]repository),
+		cache:        cache.NewObjectLRU(maxCacheSize),
```
jfontan: If I understand correctly we are using the same cache for all repositories. While this may be OK now we may have problems when partitions are used:

* I'm not really sure of the cache implementation thread safety
* Recent objects from one repo may be evicted by others and make it run slower
* Concurrent use of the same cache may be slower (locking?)

I would have one cache per repo, a pool of caches that can be evicted or one cache per partition. I think we can have one per repo for now and get back to it when partitions are in use.

repository_pool.go
```diff
@@ -178,17 +188,17 @@ func (p *RepositoryPool) AddGit(path string) error {
 
 // AddGitWithID adds a git repository to the pool. ID should be specified.
 func (p *RepositoryPool) AddGitWithID(id, path string) error {
-	return p.Add(gitRepo(id, path))
+	return p.Add(gitRepo(id, path, p.cache))
```
jfontan: I wouldn't store the cache in the git/sivaRepo. If they are different instances (one per repo) it can use lots of memory.

repository_pool.go
ajnavarro: the problem of having one cache per repo is that we will not able to know the amount of memory gitbase will use. Maybe we can use as a default value 96 MiB * number of repositories. Or apart from that, maybe we can implement an LRU with two key layers, to evict keys from a specific repository. WDYT?

kuba--: So maybe we can keep one cache in a pool but add mutex and prefix keys by repo id?

kuba--:
@jfontan - rebased. Lets go with the simplest approach, so far.

PR #466, opened 2018-09-10
Improvement/use storer options partitions

### LINKED ISSUE #441: Add a static repository option to go-git
go-git is always querying the system to search for objects. If we are sure that the repository is not going to change we can do some optimizations and skip expensive system calls.

I've tried one of these optimizations in DotGit.Object. The first time it reads all the objects into a map and instead of trying to open the file to check for them search the map. This, alongside maintaining packfiles opened, improved performance.

Repository: linux (2013), numpy, tensorflow
Number of rows: 395709
Query: `SELECT count(*) FROM commits c NATURAL JOIN ref_commits r WHERE r.ref_name = 'HEAD';`

Change from #440 : `1 row in set (20 min 43.69 sec)`
This proposal: `1 row in set (13 min 19.41 sec)`

Add a new option to git repository open to expect the repository to be static.

jfontan:
Adds storer close after repository usage. Also adds ExclusiveAccess option.

Fixes #441

### DIFF
--- a/repository_pool.go
+++ b/repository_pool.go
@@ -35,10 +35,47 @@ func NewRepository(id string, repo *git.Repository) *Repository {
 	}
 }
 
+// Close closes all opened files in the repository.
+func (r *Repository) Close() {
+	if r != nil && r.Repository != nil {
+		f, ok := r.Storer.(*filesystem.Storage)
+		if ok {
+			// The only type of error returned is "file already closed" and
+			// we don't want to do anything with it.
+			f.Close()
+		}
+	}
+}
+
 // NewRepositoryFromPath creates and initializes a new Repository structure
 // and initializes a go-git repository
 func NewRepositoryFromPath(id, path string) (*Repository, error) {
-	repo, err := git.PlainOpen(path)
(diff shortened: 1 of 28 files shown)


repository_pool.go
```diff
@@ -35,10 +35,47 @@ func NewRepository(id string, repo *git.Repository) *Repository {
...
 func NewRepositoryFromPath(id, path string) (*Repository, error) {
-	repo, err := git.PlainOpen(path)
+	op := filesystem.Options{
+		ExclusiveAccess: true,
+		KeepDescriptors: true,
+	}
+
+	var wt billy.Filesystem
+	fs := osfs.New(path)
+	f, err := fs.Stat(git.GitDirName)
```
mcarmonaa: What would happen if  `path` is a bare repository?

jfontan: If it's a bare repo then `.git` does not exist and the path bellow (`if != nil && f.IsDir()`) not taken. This means that worktree will be nil and the storer is created with the fs provided instead of chroot into `.git`.

PR #468, opened 2018-09-13
Change the notice about Spark integration

vmarkovtsev:
People get confused and think that it is already working https://github.com/src-d/datasets/issues/77#issuecomment-420935678


ajnavarro:
After implementing https://github.com/src-d/go-mysql-server/issues/359 that will be possible.

vmarkovtsev:
...which is not merged yet @ajnavarro  :)

PR #469, opened 2018-09-13
*: make SquashedTable not implement sql.Node

erizocosmico:
Depends on https://github.com/src-d/go-mysql-server/pull/352


erizocosmico:
Upgraded go-mysql-server, now everything should pass

ajnavarro:
@erizocosmico can you check codecov? appears that squash iterator is not tested anymore.

erizocosmico:
@ajnavarro it is covered, just some `Close` methods are not being called (dunno why tho) and something that's no longer used and I left there.
I'm investigating why close methods are not being called.

erizocosmico:
Turns out it's not because of this, it's that `ResolvedTable` does not close the `sql.RowIter` returned by `PartitionRows`, only the `sql.PartitionIter` 😅

erizocosmico:
Depends on https://github.com/src-d/go-mysql-server/pull/362

ajnavarro:
@erizocosmico now you can update go-mysql-server and rebase!

internal/rule/squashjoins_test.go
```diff
@@ -33,13 +34,17 @@ func TestAnalyzeSquashJoinsExchange(t *testing.T) {
 	require.NoError(err)
 
 	project, ok := result.(*plan.Project)
+	fmt.Printf("%T", result)
```
ajnavarro: leftover?

Issue #470, opened 2018-09-13
Wrong binaries provided for linux in releases


mcarmonaa:
Binaries built for linux provided in releases are not valid since they are using `libonig2` which is a library for debian distros.

If you try to run that binary in other distro (Arch in my case), you get this error:
```
./gitbase: error while loading shared libraries: libonig.so.2: cannot open shared object file: No such file or directory
```

The way to install `oniguruma` in Arch linux is:
```
pacman -S oniguruma
```

and the shared object is left in your system and that will be use when you include `oniguruma.h` in you project is:
```
$ ls -lah /usr/lib | grep libonig
lrwxrwxrwx   1 root root    16 Sep 10 08:45 libonig.so -> libonig.so.5.0.0
lrwxrwxrwx   1 root root    16 Sep 10 08:45 libonig.so.5 -> libonig.so.5.0.0
-rwxr-xr-x   1 root root  551K Sep 10 08:45 libonig.so.5.0.0

```

erizocosmico:
Did #471 fix your issue?

mcarmonaa:
@erizocosmico no, it didn't. I don't know what is the solution, maybe a comment in the releases telling about those binaries are just valid for debian-based distros?

ajnavarro:
the binaries should be valid for all linux distros. If that cannot be possible, we should go back and generate linux binaries without using oniguruma.

PR #471, opened 2018-09-13
docs: fix how to install oniguruma on linux

mcarmonaa:
Related to #470

### DIFF
--- a/docs/using-gitbase/getting-started.md
+++ b/docs/using-gitbase/getting-started.md
@@ -64,8 +64,13 @@ brew install oniguruma
 
 Linux:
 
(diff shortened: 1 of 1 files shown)


docs/using-gitbase/getting-started.md
```diff
@@ -64,9 +64,14 @@ brew install oniguruma
 
 Linux:
 
+- Debian and Debian-based distros:
```
kuba--: >  Debian-based
sounds good enough to me.

kuba--: btw. don't you need a `...-dev` version of this library to install?

PR #475, opened 2018-09-18
remove oniguruma support from prebuilt binaries

### LINKED ISSUE #470: Wrong binaries provided for linux in releases

erizocosmico:
Fixes #470


ajnavarro:
@mcarmonaa I think we should use the same binary on all the release artifacts.

PR #485, opened 2018-09-20
*: replace pilosalib with new pilosa driver

erizocosmico:
Will fail until https://github.com/src-d/go-mysql-server/pull/373 is merged and this is updated


jfontan:
As @kuba-- said in go-mysql-server issue it would be nice to add a note in the documentation to migrate from old pilosa driver. This may be done after the code is merged.

kuba--:
now, tests are failing

erizocosmico:
@kuba-- yep:

> Will fail until src-d/go-mysql-server#373 is merged and this is updated

erizocosmico:
Should be fixed now

kuba--:
Still failing, because of integration tests. For instance we have just one constructor `NewDriver`, now.

PR #486, opened 2018-09-20
internal/rule: fix reorder exchanges

### LINKED ISSUE #481: query from examples failing
This query results in error `: unknown error: unable to find field with index 10 in row of 8 columns`:
```sql
select
    files.file_path,
    ref_commits.repository_id,
    files.blob_content
FROM
    files
NATURAL JOIN
    commit_files
NATURAL JOIN
    ref_commits
WHERE
    ref_commits.ref_name = 'HEAD'
    AND ref_commits.history_index BETWEEN 0 AND 5
    AND is_binary(blob_content) = false
    AND files.file_path NOT REGEXP '^vendor.*'
    AND (
        blob_content REGEXP '(?i)facebook.*[\'\\"][0-9a-f]{32}[\'\\"]'
        OR blob_content REGEXP '(?i)twitter.*[\'\\"][0-9a-zA-Z]{35,44}[\'\\"]'
        OR blob_content REGEXP '(?i)github.*[\'\\"][0-9a-zA-Z]{35,40}[\'\\"]'
        OR blob_content REGEXP 'AKIA[0-9A-Z]{16}'
        OR blob_content REGEXP '(?i)reddit.*[\'\\"][0-9a-zA-Z]{14}[\'\\"]'
        OR blob_content REGEXP '(?i)heroku.*[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}'
        OR blob_content REGEXP '.*-----BEGIN PRIVATE KEY-----.*'
        OR blob_content REGEXP '.*-----BEGIN RSA PRIVATE KEY-----.*'
        OR blob_content REGEXP '.*-----BEGIN DSA PRIVATE KEY-----.*'
        OR blob_content REGEXP '.*-----BEGIN OPENSSH PRIVATE KEY-----.*'
    );
```

mcarmonaa:
Fixes #481

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3,6 +3,7 @@ package gitbase_test
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


mcarmonaa:
@kuba-- CI now passes

mcarmonaa:
@erizocosmico I've added an integration test

erizocosmico:
I think there might be a better solution than this for this particular problem. I've been playing with this for a bit to see if I could find a better solution inside gitbase itself but I don't think that's gonna be possible.

So, instead, what about moving the `parallelize` rule to the `AfterAll` batch where `track_process` is executed? Even if technically is an optimisation rule, we cannot ask for user rules to take this into account every time they change something.
If we move it there, squash will run before that and then we wrap the nodes normally in exchanges at the very end of the process.

WDYT @ajnavarro?

PR #496, opened 2018-09-26
internal/function: fix uast_extract returned type

mcarmonaa:
Fixes #495

Issue #495, opened 2018-09-25
invalid type: [][]string


vmarkovtsev:
I run the following query:

```sql
SELECT files.repository_id, files.file_path,
    ARRAY_LENGTH(uast_extract(UAST(
        files.blob_content,
        LANGUAGE(files.file_path, files.blob_content),
        '//*[@roleFunction]'), "internal_type")
    ) as functions
FROM files
WHERE files.file_path = "common.go"
ORDER BY functions DESC
```

with the most recent gitbase docker image. It errors with:

```
unknown error: invalid type: [][]string
```

What's funny is if I change the xpath:

```sql
SELECT files.repository_id, files.file_path,
    ARRAY_LENGTH(uast_extract(UAST(
        files.blob_content,
        LANGUAGE(files.file_path, files.blob_content),
        '//*[@roleFunction and @roleDeclaration]'), "internal_type")
    ) as functions
FROM files
WHERE files.file_path = "common.go"
ORDER BY functions DESC
```

Then I don't get this error (but I get 0 functions in go-git/common.go instead of 1 which is still wrong).

vmarkovtsev:
Another example:

```sql
SELECT files.repository_id, files.file_path,
      ARRAY_LENGTH(uast_extract(uast_xpath(UAST_mode("annotated", files.blob_content, "Go"), "//*[@roleFunction and @roleDeclaration]"), "token")) as functions
FROM files
WHERE files.file_path = "common.go"
```

mcarmonaa:
About the error `unknown error: invalid type: [][]string` a fix is coming.

On the other hand, there is some issues with the queries:
```sql
SELECT files.repository_id, files.file_path,
    ARRAY_LENGTH(uast_extract(UAST(
        files.blob_content,
        LANGUAGE(files.file_path, files.blob_content),
        '//*[@roleFunction]'), "internal_type")
    ) as functions
FROM files
WHERE files.file_path = "common.go"
ORDER BY functions DESC
```
With this query you are using `uast` and by default it returns a `semantic` uast so I don't think that the xpath query `//*[@roleFunction]` works on it.

The `uast_extract` function have special selectors for common properties (`@token`, `@role`, `@type`, `@startpos`, `@endpos`) check [the examples](https://docs.sourced.tech/gitbase/using-gitbase/examples#extracting-information-from-uast-nodes).

[...]

PR #498, opened 2018-09-27
internal/funciton: made uast_extract return a flattened array

mcarmonaa:
Related to #495 

depends on #496


internal/function/uast.go
```diff
@@ -474,25 +477,34 @@ const (
 	keyEndPos   = "@endpos"
 )
 
-func extractInfo(n *uast.Node, key string) []string {
-
-	info := []string{}
+func extractInfo(n *uast.Node, key string) []interface{} {
+	info := []interface{}{}
```
erizocosmico: s/`info := []interface{}{}`/`var info []interface{}`/

PR #499, opened 2018-09-27
Static build

jfontan:
Now the docker container is based on alpine to make it slimmer. It also
generates a new linux binary with libxml2 and oniguruma statically
linked. This build uses musl instead of glibc.

It also generates a linux static package extracted from the docker
image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,9 +38,8 @@ jobs:
 
       before_deploy:
         - make packages
-
-      after_deploy:
         - DOCKER_PUSH_LATEST=1 make docker-push
+        - make static-package
 
       deploy:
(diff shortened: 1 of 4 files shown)


ajnavarro:
Just some questions

- the static-compiled binary is not included in the docker image, right? if yes, there is some reason for that?

jfontan:
Yes, the docker image uses the static compiled binary. In fact it is generated in an alpine docker container as compiling with musl in ubuntu was problematic. Then this binary is extracted from the docker image to create the static package. That way oniguruma is enabled and we can use alpine to make it smaller:

https://hub.docker.com/r/jfontan/gitbase/tags/

PR #512, opened 2018-10-02
Dockerfile: add gitbase version to the binary

### LINKED ISSUE #508: undefined in SELECT VERSION()
Tested with the latest `v0.17.0-rc.3`:
```
> select version();
+-----------+
| VERSION() |
+-----------+
| 8.0.11-undefined |
+-----------+

```

mcarmonaa:
Closes #508

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,10 @@ RUN apk add --update libxml2-dev git make bash gcc g++ curl oniguruma-dev
 RUN go get github.com/golang/dep/...
 RUN dep ensure
 RUN cd vendor/gopkg.in/bblfsh/client-go.v2 && make dependencies
-RUN go install -v -tags oniguruma -ldflags "-linkmode external -extldflags '-static -lz'" github.com/src-d/gitbase/...
+RUN VERSION=$(git describe --exact-match --tags 2>/dev/null || \
+    echo dev-$(git rev-parse --short HEAD)$(test -n "`git status --porcelain`" && echo "-dirty" || true)); \
+    go install -v -tags oniguruma -ldflags "-linkmode external -extldflags '-static -lz' -X main.version=$VERSION" \
+    github.com/src-d/gitbase/...
 
 FROM alpine:3.8
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,10 +8,7 @@ RUN apk add --update libxml2-dev git make bash gcc g++ curl oniguruma-dev
 RUN go get github.com/golang/dep/...
 RUN dep ensure
 RUN cd vendor/gopkg.in/bblfsh/client-go.v2 && make dependencies
-RUN VERSION=$(git describe --exact-match --tags 2>/dev/null || \
-    echo dev-$(git rev-parse --short HEAD)$(test -n "`git status --porcelain`" && echo "-dirty" || true)); \
-    go install -v -tags oniguruma -ldflags "-linkmode external -extldflags '-static -lz' -X main.version=$VERSION" \
-    github.com/src-d/gitbase/...
+RUN make static-build
 
 FROM alpine:3.8
 
--- a/Makefile
+++ b/Makefile
@@ -34,3 +34,10 @@ static-package:
 	cd build && \
 	tar czvf $${PACKAGE_NAME}.tar.gz $${PACKAGE_NAME} && \
 	docker rm gitbase-temp
+
+# target used in the Dockerfile to build the static binary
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -8,7 +8,9 @@ RUN apk add --update libxml2-dev git make bash gcc g++ curl oniguruma-dev
 RUN go get github.com/golang/dep/...
 RUN dep ensure
 RUN cd vendor/gopkg.in/bblfsh/client-go.v2 && make dependencies
-RUN go install -v -tags oniguruma -ldflags "-linkmode external -extldflags '-static -lz'" github.com/src-d/gitbase/...
+RUN VERSION=dev-$(git rev-parse --short HEAD)-$(test -n "`git status --porcelain`" && echo "-dirty" || true); \
```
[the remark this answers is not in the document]
mcarmonaa: No, it isn't. Fixed :+1:

Dockerfile
```diff
@@ -8,7 +8,10 @@ RUN apk add --update libxml2-dev git make bash gcc g++ curl oniguruma-dev
 RUN go get github.com/golang/dep/...
 RUN dep ensure
 RUN cd vendor/gopkg.in/bblfsh/client-go.v2 && make dependencies
-RUN go install -v -tags oniguruma -ldflags "-linkmode external -extldflags '-static -lz'" github.com/src-d/gitbase/...
+RUN VERSION=$(git describe --exact-match --tags 2>/dev/null || \
+    echo dev-$(git rev-parse --short HEAD)$(test -n "`git status --porcelain`" && echo "-dirty" || true)); \
+    go install -v -tags oniguruma -ldflags "-linkmode external -extldflags '-static -lz' -X main.version=$VERSION" \
```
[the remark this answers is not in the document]
mcarmonaa: @ajnavarro I've changed the variables and added a new target to the `Makefile` to use it in the `Dockerfile` to build the static binary.

Makefile
```diff
@@ -34,3 +34,10 @@ static-package:
 	cd build && \
 	tar czvf $${PACKAGE_NAME}.tar.gz $${PACKAGE_NAME} && \
 	docker rm gitbase-temp
+
+# target used in the Dockerfile to build the static binary
+static-build: VERSION = $(shell git describe --exact-match --tags 2>/dev/null || dev-$(git rev-parse --short HEAD)$(test -n "`git status --porcelain`" && echo "-dirty" || true))
```
ajnavarro: You only need to call to make packages with specific LD_FLAGS and GO_BUILD_FLAGS for that case. You don't need to call to go install again and to redefine VERSION

mcarmonaa: The docker image was already using `go install` and copying the binary from `/go/bin/gitbase` from one container to another.

Do we want to use `make package` which builds several binaries for different platforms only for copy one of them?

On the other hand, `VERSION` must be overwritten because it only takes the value from the tag name when it runs in travis or drone, looking for environment variables they provide. If we don't, it will always use version names like `dev-commit`.

jfontan: I don't see an easy way to do this. Either we don't use `make docker-build` and use `docker build --build-args` to pass version information to generate the binary or the trick you're using.

Either way I don't understand how `VERSION` and `LD_FLAGS` are passed into the `go install` command.

mcarmonaa: @jfontan I'm modifying the global variables `VERSION`, `LD_FLAGS` and `GO_BUILD_ARGS` just for the scope of the target `static-build`.

[`GO_BUILD_ARGS`](https://github.com/src-d/ci/blob/v1/Makefile.main#L75) makes use of `LD_FLAGS`, and [`LD_FLAGS`](https://github.com/src-d/ci/blob/v1/Makefile.main#L69) makes use of `VERSION` so when `GO_BUILD_ARGS` is used with `go install` it gets the correct values to build gitbase inside the container.

jfontan: But isn't `GO_BUILD_FLAGS` already set at that point? AFAIK it shouldn't change when you modify `LD_FLAGS`.

mcarmonaa: `GO_BUILD_FLAGS`, `LD_FLAGS` and `VERSION` are defined as  recursively-expanded variables so the when you use `+=` you are not adding that value to an already evaluated expression, it expands later.

see https://www.gnu.org/software/make/manual/html_node/Appending.html#Appending

jfontan: I've just checked and does what it's supposed to do. Still I don't understand how :sweat_smile:

jfontan: oh! thanks, now I get it.

Issue #510, opened 2018-10-02
Cannot connect (timeout or wrong answer from the server)


kuba--:
I suppose the connection to `gitbase` *timeouts* because clients don't get the right response from the server for query:
`"SELECT @@global.max_allowed_packet"`
It's one of first queries clients send to the server.

PR #517, opened 2018-10-02
Upgrade mysql

kuba--:

Fixes: https://github.com/src-d/gitbase/issues/510

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,3 @@ main
 # CI
 .ci/
 build/
-.vscode/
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -27,3 +27,4 @@ main
 # CI
 .ci/
 build/
+.vscode/
```
erizocosmico: this should go in the user global .gitignore

kuba--: ok, done

PR #559, opened 2018-10-09
docs: add supported clients

### LINKED ISSUE #551: Tested client libraries?
Is there any documentation on which MySQL client libraries have been tested and which are suggested to be used? 

Thanks

mcarmonaa:
Closes #551

### DIFF
--- a/docs/using-gitbase/supported-clients.md
+++ b/docs/using-gitbase/supported-clients.md
@@ -1,0 +1,3 @@
+## Supported clients
+
+To see the supported MySQL clients and examples about how tho use them, take a look [here](https://github.com/src-d/go-mysql-server/blob/b1203b4cd0f4734da7552435f10b65a311226a35/SUPPORTED_CLIENTS.md).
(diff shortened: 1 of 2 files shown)


docs/using-gitbase/supported-clients.md
```diff
@@ -0,0 +1,3 @@
+## Supported clients
+
+To see the supported MySQL clients and examples about how tho use them, take a look [here](https://github.com/src-d/go-mysql-server/blob/master/SUPPORTED_CLIENTS.md).
```
erizocosmico: Instead of master, it should point to the tree of the current revision, like we don on some other parts of the doc.

mcarmonaa: I've changed to point to `https://github.com/src-d/go-mysql-server/blob/679d33772845593ce5fdf17925f49f2335bc8356/SUPPORTED_CLIENTS.md`which right now isn't a valid url, but when `go-mysql-server` dependency will be updated it will a point to the correct url.

ajnavarro:
@mcarmonaa can you upgrade this PR using the actual go-mysql-server version that is on master? thanks!

ajnavarro:
Also, you need to add a link to SUMMARY.md to this new file.

mcarmonaa:
@ajnavarro done

PR #566, opened 2018-10-14
Fixed a grammatical error

DCzajkowski:
According to [this](https://english.stackexchange.com/a/409933) comment, it's preferred to write "on a large scale" rather than "at a large scale". "at large scale" is a grammatical error.


ajnavarro:
@DCzajkowski thanks a lot!

PR #577, opened 2018-10-22
Added repository_ids in non-natural joins in the examples

### LINKED ISSUE #575: Examples are missing repository_id equality on joins
https://docs.sourced.tech/gitbase/using-gitbase/examples

All the examples not using natural joins are missing the `T1.repository_id = T2.repository_id` on the joins.

This does not matter in gitbase because it assumes it has them (incorrectly, while we're at it), but in the gitbase-spark-connector, it means it won't push down the joins. So it's better to correct it in case someone copies from these examples.

BakeRolls:
Added `repository_id`s to `JOIN .. ON` clauses and unified the indentation to four spaces (which was used before).

This would hopefully close #575.

### DIFF
--- a/docs/using-gitbase/examples.md
+++ b/docs/using-gitbase/examples.md
@@ -6,7 +6,7 @@
 SELECT refs.repository_id
 FROM refs
 NATURAL JOIN commits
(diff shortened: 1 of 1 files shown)


ajnavarro:
Thanks @BakeRolls !

PR #586, opened 2018-10-24
Update Dockerfile to reduce image size

prog1dev:
Adds `--no-install-recommends` to reduce docker image size by 18mb.


ajnavarro:
Thanks!

PR #594, opened 2018-10-31
Update LICENSE

vyaspranjal33:
added year and org name in license

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
+   Copyright 2018 source{d}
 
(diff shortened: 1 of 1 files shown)


ajnavarro:
Thanks for the contribution, but nothing to fix here, it is a template.

ajnavarro:
Related: https://github.com/src-d/github-reminder/pull/17#issuecomment-432286706

smola:
As noted in https://github.com/src-d/github-reminder/pull/17#issuecomment-432286706 there is nothing to fix in the `LICENSE` file. That text is part of the instructions about adding a notice to the source code, it's not the notice itself.

You can see projects by the Apache Foundation itself, where they just leave the Apache License 2.0 text verbatim, including the appendix: https://github.com/apache/spark/blob/master/LICENSE

PR #596, opened 2018-11-01
Refactored Dockerfile

Lerentis:
Hey there,

i saw that your Dockerfile depended on Debian. I refactored it to depend on Alpine which is much smaller. This should not change anything functional, it just decreases the image size

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,7 @@ ENV TINI_VERSION v0.18.0
 ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-amd64 /tini
 RUN chmod +x /tini
 
-RUN echo 'http://dl-cdn.alpinelinux.org/alpine/v3.1/main' >> /etc/apk/repositories && \ 
-	apk --no-cache add 'libxml2==2.9.2-r2' git oniguruma libc6-compat &&  \
-	apk clean
+RUN apk --no-cache add libxml2 git oniguruma libc6-compat
 
 ENTRYPOINT ["/tini", "--"]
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ $(MAKEFILE):
 
 # we still need to do this for windows
 bblfsh-client:
-	cd vendor/gopkg.in/bblfsh/client-go.v2 && make dependencies
+	cd vendor/gopkg.in/bblfsh/client-go.v3 && make dependencies
```
jfontan: Why the change in client-go version? If I am not mistaken we are still using v2.

erizocosmico: Nope, we're using v3 in master

Dockerfile
```diff
@@ -8,15 +8,13 @@ ENV GITBASE_PASSWORD=""
...
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-amd64 /tini
 RUN chmod +x /tini
 
-RUN apt-get update \
-  && apt-get -y --no-install-recommends install libxml2 git \
-  && apt-get clean \
-  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
+RUN echo 'http://dl-cdn.alpinelinux.org/alpine/v3.1/main' >> /etc/apk/repositories && \
```
erizocosmico: shouldn't this be `v3.8` to match the alpine version in use?

Lerentis:
I was not sure about the version of libxml so I forced it with this repository. If the version of libxml does not matter I can remove the whole repository there

-----Original Message-----

erizocosmico:
I would say if it compiles and works correctly, there's no libxml version requirement afaik.

Lerentis:
I removed the version pin and with that the extra repository is not needed anymore

Issue #547, opened 2018-10-05
Soft links not supported


campoy:
I tried starting a gitbase server pointing to a soft link `~/new_repos` which pointed to a directory `/mnt/disks/sdb/new_repos` which contains a bunch of cloned repositories.

When running `gitbase -d ~/new_repos` no repos were found, while running `gitbase -d /mnt/disks/sdb/new_repos` did find all of the repositories.

BakeRolls:
Hi, since `filepath.Walk` doesn't follow symlinks, we could check the provided directories and optionally resolve them. Deeper links should probably still be ignored.

Please have a look at [my branch](https://github.com/src-d/gitbase/compare/master...BakeRolls:server-follow-symlink) for a very small implementation, if this solution would work for you.

campoy:
Just tried and unfortunately it doesn't work.
If you send a PR I'm happy to review and try to see why it's not.

ajnavarro:
@campoy did you test symlinks with the latest beta version?

campoy:
Maybe I'm doing it wrong ... but it doesn't seem to be working.

I created a directory with a soft link pointing to a directory containing a bunch of repos, none of the repos were analyzed.

kuba--:
@campoy - actually I cannot reproduce it. Which one is your case?
### Case 1 (symlink to folder with repos)
```sh
# download/clone some repositories to repos directory e.g.:
$ pga list -l python -f json | head -n 10 | jq -r '.sivaFilenames[]' | pga get -i -o repos

# create symlink to repos directory
$ ln -s repos repos-link

# start gitbase
$ ./gitbase server -d repos-link
```

### Case 2 (with symlink as subdirectory)
```sh
# download/clone somerepositories to repos directory e.g.:
$ pga list -l python -f json | head -n 10 | jq -r '.sivaFilenames[]' | pga get -i -o repos

# create another directory
$ mkdir repos2
$ cd repo2

# create symlink to repos subdirectory
$ ln -s ../repos/siva siva
$ cd ..

# start gitbase
$ ./gitbase server -d repos2
```

*Case 1* should work (because we resolve symlinks just for the top level), but when we traverse subdirectories we don't resolve any links, so *Case 2* doesn't work.

campoy:
I think my case is actually somewhere in between:

There's a folder `repos` that contains a bunch of repositories.
Then I have another directory `home` that contains a `repos` soft link to the `repos` directory above.

I can try it again if needed, since it's been a while.

[...]

PR #597, opened 2018-11-03
Server follow symlinks when passed as an argument

BakeRolls:
This PR tries to close issue #547. The server should follow symlinks passed as an argument. For example, if I have a directory `~/Desktop/repos` and a softlink `~/Desktop/repos-link -> ~/Desktop/repos`, running

```
$ go run main.go server -d ~/Desktop/repos/
```

and 

```
go run main.go server -d ~/Desktop/repos-link/
```

should behave the same.

### DIFF
--- a/cmd/gitbase/command/server.go
+++ b/cmd/gitbase/command/server.go
@@ -273,6 +273,17 @@ func (c *Server) addMatch(match string) error {
 		return err
 	}
 
+	rooti, err := os.Lstat(root)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


arfat392:
Love

BakeRolls:
I could only think of very basic tests that check if adding a softlink fails. Do you have ideas how to improve on it?

BakeRolls:
And thank you for the reminder, @arfat392!

PR #611, opened 2018-11-07
function: make max bblfsh blob size configurable

### LINKED ISSUE #604: UASTs functions: configurable limit on files sent to bblfsh.
Add a configuration param to be able to configure the size of files sent to bblfsh.

erizocosmico:
Closes #604


docs/using-gitbase/configuration.md
```diff
@@ -16,6 +16,7 @@
 | `GITBASE_CACHESIZE_MB`       | size of the cache for git objects specified as MB                                  |
 | `GITBASE_CONNECTION_TIMEOUT` | timeout in seconds used for client connections on write and reads. No timeout by default.     |
 | `GITBASE_USER_FILE`          | JSON file with user credentials                                                    |
+| `GITBASE_MAX_UAST_BLOB_SIZE`          | Max size of blobs to send to be parsed by bblfsh. No limit by default.                                                    |
```
kuba--: I have generic question - do people really know (should know) anything about bblfsh and/or UAST, when using gitbase?

internal/function/uast.go
```diff
@@ -20,14 +20,18 @@ import (
...
-	uastCacheSize        = "GITBASE_UAST_CACHE_SIZE"
+	uastCacheSizeKey     = "GITBASE_UAST_CACHE_SIZE"
 	defaultUASTCacheSize = 10000
+
+	uastMaxBlobSizeKey     = "GITBASE_MAX_UAST_BLOB_SIZE"
+	defaultUASTMaxBlobSize = -1
 )
 
 var uastCache *lru.Cache
+var uastMaxBlobSize int
```
kuba--: we should specify _unit_, I think it make sense to have it in MB. But if we stick to bytes then `int` may be not good enough.

erizocosmico: `len(bytes)` is an int. You literally can't have a slice bigger than that, so no point in using any other type

docs/using-gitbase/configuration.md
erizocosmico: I guess not using gitbase, but yes when using uasts in gitbase

internal/function/uast.go
```diff
@@ -21,14 +21,18 @@ import (
 )
 
 const (
-	uastCacheSize        = "GITBASE_UAST_CACHE_SIZE"
+	uastCacheSizeKey     = "GITBASE_UAST_CACHE_SIZE"
 	defaultUASTCacheSize = 10000
+
+	uastMaxBlobSizeKey     = "GITBASE_MAX_UAST_BLOB_SIZE"
+	defaultUASTMaxBlobSize = -1
```
ajnavarro: we should use the same default as in bblfsh

erizocosmico:
Done

PR #612, opened 2018-11-07
*: make UAST functions fail on connection errors

### LINKED ISSUE #603: UASTs functions: add as warnings only parsing errors on bblfsh functions.
Connection errors (and other kind of unexpected errors) after retries should make the query fail.

erizocosmico:
Fixes #603

This makes it fail on connection errors. For bblfsh bugs, though, it's not that easy. Bblfsh parse returns an error, we cannot distinguish between bugs and invalid code sent.


ajnavarro:
@erizocosmico on v3 bblfsh protocol you have kind of errors for that cases: https://github.com/bblfsh/client-go/blob/master/request.go#L19

jfontan:
@erizocosmico they've recently added error types for parsing. Maybe with this it is possible:

https://github.com/bblfsh/sdk/pull/321

erizocosmico:
@jfontan nice! that's exactly what we need

erizocosmico:
Updated to include other driver non-syntax errors

Issue #608, opened 2018-11-06
Scheme diagram image is outdated


dpordomingo:
`remotes` table has no `remote_name`
`commits` table has no `commit_message`
`ref_commits` table says `index` instead of `history_index`

PR #620, opened 2018-11-16
Update docs (schema diagram)

kuba--:

Closes https://github.com/src-d/gitbase/issues/608


kuba--:
FYI: I've added binary assets to git-lfs

PR #629, opened 2018-11-27
docs: add explanation about using explain to assess performance issues

### LINKED ISSUE #624: explain should be documented on "Optimize queries" doc
Explain is not ... explained 🥁  in https://github.com/src-d/gitbase/blob/master/docs/using-gitbase/optimize-queries.md

We should have some basic explanation of the command and how to interpret the results according to the explained concepts on indexes, squashed tables, etc.

erizocosmico:
Closes #624

### DIFF
--- a/docs/using-gitbase/optimize-queries.md
+++ b/docs/using-gitbase/optimize-queries.md
@@ -6,6 +6,58 @@ There are two ways to optimize a gitbase query:
 - Create an index for some parts.
 - Making sure the joined tables are squashed.
 
+## Assessing performance bottlenecks
+
+To assess if there is a performance bottleneck you might want to inspect the execution tree of the query. This is also very helpful when reporting performance issues on gitbase.
+
+The output from an `EXPLAIN` query is represented as a tree and shows how the query is actually evaluated.
+You can do that using the following query:
+
+```sql
+EXPLAIN FORMAT=TREE <SQL QUERY TO EXPLAIN>
+```
+
+For example, the given query:
+
+```sql
+EXPLAIN FORMAT=TREE
+    SELECT * FROM refs
+    NATURAL JOIN ref_commits
+    WHERE ref_commits.history_index = 0
+```
+
+Will output something like this:
+
+```
++-----------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


docs/using-gitbase/optimize-queries.md
```diff
@@ -6,6 +6,57 @@ There are two ways to optimize a gitbase query:
 - Create an index for some parts.
 - Making sure the joined tables are squashed.
 
+## Assessing performance bottlenecks
+
+To assess if there is a performance bottleneck you might want to inspect the execution tree of the query. This is also very helpful when reporting performance issues on gitbase.
+
+The query tree can be printed with the following query:
```
kuba--: Maybe....?

>The output from EXPLAIN QUERY is represented as a tree and shows how the query is actually evaluated.
> For example, the following query:
> ...

docs/using-gitbase/optimize-queries.md
```diff
@@ -6,6 +6,57 @@ There are two ways to optimize a gitbase query:
...
+|          └─ ref_commits.history_index = 0                                               |
++-----------------------------------------------------------------------------------------+
+15 rows in set (0.00 sec)
+```
+
+#### Detecting performance issues in the query tree
+
+Some performance issues might not be obvious, but there are a few that really stand out by just looking at the query tree.
+
+- Joins not squashed. If you performed some joins between tables and instead of a `SquashedTable` node you see `Join` and `Table` nodes, it means the joins were not successfully squashed. There is a more detailed explanation about this in later sections of this document.
```
kuba--: next instead of later?
> ...in next sections of this document

docs/using-gitbase/optimize-queries.md
```diff
@@ -6,6 +6,57 @@ There are two ways to optimize a gitbase query:
...
+- Indexes not used. If you can't see the indexes in your table nodes, it means somehow those indexes are not being used by the table. There is a more detailed explanation about this in later sections of this document.
```
kuba--: s/later/ next, below, ... /

erizocosmico:
Done @kuba--

PR #632, opened 2018-11-28
add warns to session on uast functions

### LINKED ISSUE #630: Implement Session.Warn on places that can be useful.
We can call to Session.Warn on gitbase Nodes that can give useful information to the user. From the top of my head, we can implement it on UAST UDFs.

erizocosmico:
Closes #630


ajnavarro:
@kuba-- Regarding this:

> Maybe it's just a one-time case. On the other hand -
does it make sense to start thinking about ctxlogrus where all warnings also will go to ctx.Warn?

Maybe not all the log warnings should go to ctx.Warn, maybe just for now is better to have them separated.

Issue #638, opened 2018-11-28
Enable Gitbase->Bblfsh request tracing


dennwc:
We will soon release a bblfshd with OpenTracing enabled, and it will be nice to propagate Gitbase traces to bblfshd, as was shown during the demo.

There are a few requirements to make this work:
- [ ] SDK v2.9.x.
- [ ] Go client v3.2.x.
- [ ] Initialize `opentracing.GlobalTracer` during startup (see `cfg.InitGlobalTracer`).

I also propose to revisist how tracer is used in Gitbase. Maybe it's possible to use `GlobalTracer` as well? For now I see no reason not to.

dennwc:
As a note, once the decision about `GlobalTracer` is made, I will be happy to push a PR that implements it.

PR #642, opened 2018-12-03
Use global open tracer

kuba--:
This PR Closes https://github.com/src-d/gitbase/issues/638

- `dep ensure -v -update gopkg.in/bblfsh/sdk.v1 gopkg.in/bblfsh/sdk.v2 gopkg.in/bblfsh/client-go.v3`
- replace own gitbase tracer by global open tracer


cmd/gitbase/command/server.go
```diff
@@ -146,23 +146,19 @@ func (c *Server) Execute(args []string) error {
...
-
+		closer, err := cfg.InitGlobalTracer(cfg.ServiceName, config.Logger(logger))
 		if err != nil {
-			logrus.WithField("error", err).Fatal("unable to initialize tracer")
+			logrus.WithField("error", err).Fatal("unable to initialize global tracer")
 			return err
 		}
 
-		tracer = t
+		tracer = opentracing.GlobalTracer()
```
dennwc: Do you need the tracer in this case? Any library can call `opentracing.GlobalTracer()` to get it. I think it's probably a single package that generates a root context for requests.

kuba--: yes, but we pass it to the go-mysql-server, so thanks to this we can use global (or any other tracer) in gitbase without changing anything in mysql package.

ajnavarro:
@kuba-- shall we change the needed bblfsh version on documentation? (At least here: https://github.com/src-d/gitbase/blob/master/docs/using-gitbase/getting-started.md)

PR #650, opened 2018-12-05
gitbase: show pushed down properties to tables in explain

### LINKED ISSUE #647: can't see if an index is used using EXPLAIN
Regular mysql add extra column with something like `Using index condition` in explain. But I don't see it with gitbase. Though indexes where applied (because the query became much faster).

erizocosmico:
Fixes #647 

Turns out it was just implemented for squashed tables


erizocosmico:
Fixed

Issue #658, opened 2018-12-12
Add supported languages list from enry on gitbase documentation


ajnavarro:
Issue extracted from https://github.com/src-d/gitbase/issues/443

ajnavarro:
https://github.com/bblfsh/documentation/issues/191

smola:
The issue in the linked issue is not in bblfsh:
`LANGUAGE(t.tree_entry_name, b.blob_content) = 'go'`

That's enry. You should link to https://github.com/src-d/enry/blob/master/data/alias.go or github/linguist for valid values. That could be a note in your documentation about the `LANGUAGE` function.

smola:
AFAIK `go` would be a valid value for bblfsh, the problem here is comparing it with `LANGUAGE` result.

PR #659, opened 2018-12-12
Add list of currently supported languages

kuba--:

Closes https://github.com/src-d/gitbase/issues/658


ajnavarro:
Shall we wait til https://github.com/bblfsh/documentation/issues/191 is resolved? and just point to that documentation?

kuba--:
@ajnavarro @smola - how about now (I've updated a list, grouped by type and added links to enry and linguist)

smola:
This needs clafication. `LANGUAGE` function supports everything supported by src-d/enry. `UAST` supports whatever bblfsh has drivers for (if they are installed).

kuba--:
@smola PTAL

docs/using-gitbase/supported-languages.md
```diff
@@ -0,0 +1,517 @@
...
+- Org
+- Pod
+- Pod 6
+- RDoc
+- RMarkdown
+- Text
+- Textile
+- reStructuredText
+
+More details about aliases, groups, extensions, etc. you can find in [enry's repo](https://github.com/src-d/enry/blob/master/data/alias.go),
```
smola: Maybe all these explanations should come first, and the language list last? Because since it's very long, people may miss the distinction between enry-supported and bblfsh-supported.

kuba--: Maybe I'll leave just links and remove the list of all languages - WDYT?

kuba--: Ok, updated and rebased

PR #664, opened 2018-12-18
add e2e tests against the built binary

### LINKED ISSUE #660: undefined in SELECT VERSION()
Same as #508.
The docker image for 0.18.0 returns `8.0.11-undefined` to `SELECT VERSION()`.

erizocosmico:
Partially f_i_xes #660


erizocosmico:
Done @kuba--

jfontan:
At some point we should have only one generated binary. I believe this is only testing the one generated outside of docker.

PR #668, opened 2019-01-04
regression: add Jenkinsfile

jfontan:
Only runs on `master` branch and it checks SCM twice a day. Uses the queries from `_testdata/regression.yml`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,54 @@
+pipeline {
+  agent {
+    kubernetes {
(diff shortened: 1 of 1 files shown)


smola:
> AFAIK there is no way of merging parts hardcoded with the steps specific to projects, for example taking out the kubernetes part from this file.

As discussed offline, there is nothing here that we need to keep secret. Other than that, we'll eventually refactor these Jenkinsfile to use a shared library, to reduce boilerplate.

PR #670, opened 2019-01-11
gitbase: implement table checksums

### LINKED ISSUE #661: Can we avoid "index read failed: EOF"?
Whenever a dataset changes the indices on it becomes useless and trying to read them causes the server to respond with `index read failed: EOF`.

I'd like to avoid these issues and simply ignore the failing indexes rather than failing the whole request.

erizocosmico:
Closes #661 

Depends on https://github.com/src-d/go-mysql-server/pull/583

Tests will fail until that PR is merged and this is updated.

Basically, this implements the new `sql.Checksumable` interface in each table so the engine can perform the checksums on the repos.

How is the checksum implemented?
With a hash built using the repo ID, all the reference names and the hashes they point to and the checksum of all their packfiles.

If there's a single thing of those that changed, the checksum will be different.

Bad part of this? Pool is needed again in the tables :/ The checksums are checked as part of the init phase of the engine, and there is no context to get the repos from.

### DIFF
--- a/checksum.go
+++ b/checksum.go
@@ -1,0 +1,116 @@
+package gitbase
+
+import (
+	"bytes"
+	"crypto/sha1"
+	"encoding/base64"
+	"io"
+
+	git "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+type checksumable struct {
+	pool *RepositoryPool
+}
+
+func (c *checksumable) Checksum() (string, error) {
+	hash := sha1.New()
+	for _, id := range c.pool.idOrder {
+		repo := c.pool.repositories[id]
+		hash.Write([]byte(id))
+
+		bytes, err := readChecksum(repo)
+		if err != nil {
+			return "", err
+		}
+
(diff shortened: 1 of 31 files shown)


checksum.go
```diff
@@ -0,0 +1,116 @@
...
+
+	head, err := repo.Head()
+	if err != nil && err != plumbing.ErrReferenceNotFound {
+		return nil, err
+	} else {
+		buf.WriteString("HEAD")
+		buf.WriteString(head.Hash().String())
+	}
+
+	refs, err := repo.References()
```
kuba--: It may not work. I came across situation where ref iterator returns refs in random order - we have to double check it.

kuba--:
`table.go:12: undefined: sql.Checksumable`

erizocosmico:
Yes, until the pr on go mysql server is merged I cannot update it here. It’s not going to compile until then.

erizocosmico:
go-mysql-server PR was already merged, so just updated the dependency and tests should be working fine now.

erizocosmico:
Fixed

Issue #673, opened 2019-01-14
sql.IndexValueIter is not used via indexLookup


kuba--:
Some regression bug happened and we stopped iterate through indexed values!

I created an index:
`CREATE INDEX a ON commit_files USING pilosa (commit_hash);`

and then query:
```
SELECT commit_hash, file_path, blob_hash
FROM commit_files
WHERE commit_hash = 'd5b04a971262372a1531119d595cd4b04db2949a';
```

`Index.Get` function was called:
```
func (idx *pilosaIndex) Get(keys ...interface{}) (sql.IndexLookup, error)
```
but after that no one asked for `sql.IndexValueIter`. In other words the function:
```
func (l *indexLookup) Values(p sql.Partition) (sql.IndexValueIter, error) 
```
_wasn't called_. 
Returned result was correct.

PR #674, opened 2019-01-15
Set index for commitFilesRowIter.

kuba--:

Fixes https://github.com/src-d/gitbase/issues/673

Any good ideas how to test it?

### DIFF
--- a/commit_files.go
+++ b/commit_files.go
@@ -102,8 +102,16 @@ func (t *commitFilesTable) PartitionRows(
 				return nil, err
 			}
 
+			var index sql.IndexValueIter
+			if t.index != nil {
+				if index, err = t.index.Values(p); err != nil {
+					return nil, err
+				}
+			}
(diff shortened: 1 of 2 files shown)


ajnavarro:
@kuba-- I would say for testing, make an integration test with an `EXPLAIN` statement after creating the index and see if it is using it.

kuba--:
@ajnavarro - first of all we have to fix `EXPLAIN`. The second (what this case showed) is that even if `EXPLAIN` shows it will use indexes (because `Get(key)` was called), at the end lookup and value iterator wasn't used. So now, we have to test that `PartitionRows` calls `IndexValueIter`.

kuba--:
@ajnavarro - I've added simple test for commit_files PartitionRows with Index. 
After merge all new fixes (with explain etc.), we can add more tests based on `EXPLAIN` output.

PR #685, opened 2019-01-24
Set log level

### LINKED ISSUE #677: Let turn off logging
Even if you set a logging level to Error or Fatal in gitbase Server.Execute (or even in main) it still prints out INFO+ logs on stdout.

jfontan:
You can disable errors setting the log level to `fatal`.

Fixes #677

### DIFF
--- a/cmd/gitbase/command/server.go
+++ b/cmd/gitbase/command/server.go
@@ -108,7 +108,10 @@ func NewDatabaseEngine(
 func (c *Server) Execute(args []string) error {
 	if c.Verbose {
 		logrus.SetLevel(logrus.DebugLevel)
-	} else {
+	}
+
+	// info is the default log level
+	if c.LogLevel != "info" {
 		level, err := logrus.ParseLevel(c.LogLevel)
 		if err != nil {
 			return fmt.Errorf("cannot parse log level: %s", err.Error())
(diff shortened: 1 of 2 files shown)


cmd/gitbase/command/server.go
```diff
@@ -107,6 +108,12 @@ func NewDatabaseEngine(
 func (c *Server) Execute(args []string) error {
 	if c.Verbose {
 		logrus.SetLevel(logrus.DebugLevel)
```
kuba--: so maybe we don't need verbose anymore. it kind of duplicates log-level.
so by default we'll have info, but if you need more set up to debug.

jfontan: It does but I also didn't want to break anything that used `-v`. Maybe we can deprecate it and hide the option from the help but continue supporting it.

kuba--: Of course, but right now -v and --log-level are exclusive. In other words, if you specify `-v` then `--log-level` is ignored. Maybe we keep the old `-v` but let --log-level to overwrite it (so move it outside `else`) - WDYT?

jfontan: I'll change it to do as you say.

jfontan: Changed the behavior and updated documentation.

PR #688, opened 2019-01-30
vendor: upgrade go-mysql-server and go-git

erizocosmico:


erizocosmico:
I think there's a regression in go-git. Now we're getting object not found on a test that shouldn't give one. I'm trying to isolate this into a reproduction case so we can file an issue but I can't reproduce it (except running the tests).

jfontan:
The fix is in src-d/go-git#1056

erizocosmico:
Upgraded and fixed :tada:

Issue #617, opened 2018-11-14
Interesting performance issue counting commits


campoy:
I cloned https://github.com/kubernetes/kubernetes in order to count how many commits I can find.

```bash
$ git clone https://github.com/kubernetes/kubernetes
$ cd kubernetes
$ git log | grep "^commit " |  wc -l
72071
```

Counting the commits accessible from `HEAD` in this way takes around 2 seconds on my MacBook pro.

Next step is doing the same thing with gitbase.

```
$ time srcd sql "select count(*) from commits"
+----------+
| COUNT(*) |
+----------+
|    79991 |
+----------+
srcd sql "select count(*) from  commits;"  0.02s user 0.03s system 0% cpu 18.415 total
```

Lastly, I tried to see whether adding cores would help. Running on a GCP instance with 96 cores and way more RAM that we need, the analysis 

```
+----------+
| COUNT(*) |
+----------+
|    79991 |
+----------+
1 row in set (22.90 sec)
```

It takes longer than before!  I assumed it was before my laptop has an SSD, while this instance was using a HD ... so I tried storing the dataset (just Kubernetes) in RAM. The result was interesting ... as in it took longer than before!

```
+----------+
| COUNT(*) |
+----------+
|    79991 |
+----------+
1 row in set (23.72 sec)
```

I have no idea why this is, but it goes completely against my expectations.

jfontan:
Right now each partition is one repository so increasing the number of cores does not help. Also these many core systems usually have cores slightly slower that top of the line laptops.

ajnavarro:
Not important but with that command you are counting only commits on master history.
The correct command is:

```
$ time git log --format=oneline --all | wc -l
80028

real	0m1.063s
user	0m1.064s
sys	0m0.200s
```

Comparing the same script, but instead of using git, using go-git:
```go
package main

import (
    "fmt"
    "io"

    "gopkg.in/src-d/go-git.v4"
    . "gopkg.in/src-d/go-git.v4/_examples"
)

func main() {
    r, err := git.PlainOpen(".git")
    CheckIfError(err)

    iter, err := r.CommitObjects()
    CheckIfError(err)
    for {
        c, err := iter.Next()
        if err == io.EOF {
            break
        }

        CheckIfError(err)
        fmt.Println(c.Hash.String())
    }

    iter.Close()
}
```
Result:

```
$ time /tmp/list-commits | wc -l
80028

real	0m20.113s
user	0m19.868s
sys	0m2.220s
```
[...]

PR #698, opened 2019-02-13
 Perf: Traverse commit history instead of reading the packfile

ajnavarro:
To avoid show orphan commits and improve performance on commits table, instead of reading all the objects from each repository and discard commits, we can execute from go-git a git log --all command and traverse commit history, avoiding read a lot of data from disk.

It closes https://github.com/src-d/gitbase/issues/617


kuba--:
Let's fix tests, first.

commits.go
```diff
@@ -161,7 +162,10 @@ func (i *commitIter) init() error {
 	if len(i.hashes) > 0 {
 		i.iter, err = NewCommitsByHashIter(i.repo, i.hashes)
 	} else {
-		i.iter, err = i.repo.CommitObjects()
+		i.iter, err =
+			i.repo.Log(&git.LogOptions{
+				All: true,
```
kuba--: Maybe we should force order by time (WDYT)?
By default  pre-order will be used, which I think works slightly different than native git implementation.

ajnavarro:
@kuba-- @erizocosmico tests finally passing

commits.go
ajnavarro: but here we don't want to mimic 100% git log output, we want to reduce unnecessary disk reads. Ordering commits will not give us any performance improvement.

ajnavarro:
Some numbers using gitbase-regression with complexity=2:
```
ID      | remote:master   | local:perf/list-only-reachable-objects 
 query00 | 4.012231112s    | 1.482114222s 
 query01 | 115.760157ms    | 114.007921ms 
 query02 | 1.022439327s    | 1.008603787s 
 query03 | 937.115316ms    | 931.223158ms 
 query04 | 3.701316738s    | 1.361501136s 
 query05 | 1.377949809s    | 1.298451385s 
 query06 | 11m14.10874939s | 10m27.993923365s 
 query07 | 4.14140327s     | 1.388045549s 
 query08 | 4.193708919s    | 1.375377416s 
 query09 | 4.168560944s    | 1.395838104s 
 query10 | 6.991478009s    | 3.85651492s 
 query11 | 10m12.68998881s | 10m6.052184916s 
 query12 | 4.961075222s    | 4.957561797s 
 query13 | 116.180972ms    | 110.658067ms 
```

Issue #725, opened 2019-03-11
Limit directive takes a lot on commits table


dpordomingo:
Querying one commit takes a lot,
```SQL
select commit_author_name from commits limit 1
```

"exactly" the same than counting all of them:
```SQL
select count(*) from commits
```

If you run it from a db backing only one single repo, that time is "exactly" the same that the time needed by `go-git` to traverse the whole history of that repo from `HEAD`. (using, for example, this `go-git` [command](https://gist.github.com/dpordomingo/25722d501803208d623b749115d2f20e))

### How to reproduce;

I tried with `linux` repository doing the following:

```shell
gitbase server --port=3306 \
    --directories=$GOPATH/src/github.com/torvalds/linux \
    --index=/tmp/gitbase
```
log:
<details><pre>
INFO[0000] squash tables rule is enabled
INFO[0000] server started and listening on localhost:3306

INFO[0005] NewConnection: client 1
ERRO[0005] Error parsing auth server config: unexpected end of JSON input
INFO[0005] audit trail                  action=authentication address="127.0.0.1:39122" success=true system=audit user=root
INFO[0005] audit trail                  action=authorization address="127.0.0.1:39122" connection_id=1 permission=read pid=1 query="select @@version_comment limit 1" success=true system=audit user=root
INFO[0005] audit trail                  action=query address="127.0.0.1:39122" connection_id=1 duration=13.467911ms pid=1 query="select @@version_comment limit 1" success=true system=audit user=root
INFO[0005] audit trail                  action=authorization address="127.0.0.1:39122" connection_id=1 permission=read pid=2 query="select count(*) from commits" success=true system=audit user=root

INFO[0019] NewConnection: client 2
INFO[0019] audit trail                  action=authentication address="127.0.0.1:39124" success=true system=audit user=root
INFO[0019] audit trail                  action=authorization address="127.0.0.1:39124" connection_id=2 permission=read pid=3 query="select @@version_comment limit 1" success=true system=audit user=root
INFO[0019] audit trail                  action=query address="127.0.0.1:39124" connection_id=2 duration="370.886µs" pid=3 query="select @@version_comment limit 1" success=true system=audit user=root
INFO[0019] audit trail                  action=authorization address="127.0.0.1:39124" connection_id=2 permission=read pid=4 query="SHOW PROCESSLIST" success=true system=audit user=root
[...]

PR #726, opened 2019-03-12
no init git-log all

kuba--:

Closes https://github.com/src-d/gitbase/issues/725

I'm not sure if it's a best use case (I tried different approaches), but I'm not sure if it can be optimized without optimization options in go-git.
So far, I've removed `Log:All` from commit's init iterator.

### DIFF
--- a/commits.go
+++ b/commits.go
@@ -163,9 +163,7 @@ func (i *commitIter) init() error {
 		i.iter, err = NewCommitsByHashIter(i.repo, i.hashes)
 	} else {
 		i.iter, err =
-			i.repo.Log(&git.LogOptions{
-				All: true,
-			})
+			i.repo.Log(&git.LogOptions{})
 	}
 
 	return err


kuba--:
There is something wrong with this simple fix. For some repos I get:
`ERROR: 1105 (HY000): unknown error: reference not found`

commits.go
```diff
@@ -163,9 +163,7 @@ func (i *commitIter) init() error {
 		i.iter, err = NewCommitsByHashIter(i.repo, i.hashes)
 	} else {
 		i.iter, err =
-			i.repo.Log(&git.LogOptions{
-				All: true,
-			})
+			i.repo.Log(&git.LogOptions{})
```
dpordomingo: I'm not sure if it would change the current behavior, but I understood that `git.LogOptions{All:true}` will return all commits of the repo, no matter where they're or from where they're reachable.
Using `git.Log` with no `All:true` will return only commits being reachable from `HEAD`, what can be not enough for gitbase purpose.

kuba--:
I closed this PR. It should be a `git log --all`. In other words the algorithm for `log --all` should be improved in `go-git`

commits.go
kuba--: Without `all:true` we'll get an empty result for siva files.

PR #727, opened 2019-03-12
Add a nodeSelector tag for Jenkins on k8s

meyskens:
This adds a `nodeSelector` for Jenkins, as it inherits from perfomance it doesn't use the one set in the settings.


meyskens:
Thanks for the quick merge! Too bad it doesn't seem to work

Issue #724, opened 2019-03-11
Docs: Update query examples to use the latest and clearest syntax


ajnavarro:
Some of the examples are really old and they are not using the latest syntax like NATURAL JOIN.

Rewrite queries to use the latest syntax and be sure that they are performant enough using EXPLAIN with all of them.

PR #730, opened 2019-03-13
Update examples

kuba--:

Closes https://github.com/src-d/gitbase/issues/724


docs/using-gitbase/examples.md
```diff
@@ -140,13 +132,17 @@ Also, you can take a look to all the UDFs and their signatures in the [functions
 ## Retrieving UASTs with the UDF `uast`
 
 ```sql
-SELECT file_path, uast(blob_content, language(file_path)) FROM files;
+SELECT file_path,
+       uast(blob_content, language(file_path))
```
ajnavarro: we talked about having better examples with UAST, because this is not giving any useful information to the user. Could we change this to get some specific tokens from HEAD commits, per example (because this query is really heavy too)?

docs/using-gitbase/examples.md
```diff
@@ -140,13 +132,17 @@ Also, you can take a look to all the UDFs and their signatures in the [functions
...
+FROM files;
 ```
 
 This function allows you to directly filter the retrieved UAST by performing a XPATH query on it:
 
 ```sql
-SELECT file_path, uast(blob_content, language(file_path), "//uast:FunctionGroup") FROM files;
+SELECT file_path,
```
ajnavarro: The same, maybe is worth it to filter by HEAD commits only and get the tokens using UAST_EXTRACT udf.

kuba--: I could not get anything with `uast` what makes sense and is usable ;-(
Some queries I've just wrapped by `uast_extract`. 
It even proofs me that everything around querying `uast` is messed up (all samples in engine etc.). Returning binary data doesn't make sense to me and it's not intuitive.

kuba--: PTAL on update.

docs/using-gitbase/examples.md
```diff
@@ -6,105 +6,97 @@
...
+       repo_id,
+       committer_email
+FROM
+  (SELECT MONTH(committer_when) AS MONTH,
+          repository_id AS repo_id,
+          committer_email
+   FROM ref_commits
+   NATURAL JOIN commits
+   WHERE ref_name = 'HEAD'
+     AND YEAR(committer_when) = 2018 ) AS t
```
ajnavarro: could group by year to make this query more general?

kuba--: @ajnavarro: I've updated - take a look

docs/using-gitbase/examples.md
ajnavarro: @kuba-- we were returning UAST data as json, but it was really slow compared with this new serialization.

PR #731, opened 2019-03-13
Update enry dependency to 1.7.2

kuba--:

This PR improves accuracy for `language` function. So far, we used only `enry.GetLanguage` what for `.h` files returns an empty string. The PR tries (4 times) to get a language by cascade of functions (`GetLanguage`, `GetLanguageByContent`, `GetLanguageByFilename`, `GetLanguageByExtension`).


internal/function/language.go
```diff
@@ -161,6 +161,22 @@ func (f *Language) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
...
+		return l
+	}
+	if l, _ := enry.GetLanguageByFilename(filename); l != "" {
+		return l
+	}
+	if l, _ := enry.GetLanguageByExtension(filename); l != "" {
+		return l
+	}
+	return ""
+}
```
erizocosmico: This is already done by `enry.GetLanguage` as far as I know.

`enry.GetLanguage` calls `enry.GetLanguages` (https://github.com/src-d/enry/blob/master/common.go#L121) which calls all these strategies: https://github.com/src-d/enry/blob/master/common.go#L20-L27

kuba--: ok, so I'll change it and just call `GetLanguages`, because so far `GetLanguage` doesn't check if returned string is empty.
My intention was to add support for `.h` (c header files) which is missed.

kuba--: There is a bug in enry: https://github.com/src-d/enry/blob/master/common.go#L139
It populates a list of candidates but returns an empty slice of languages.

kuba--:
First it needs to be fixed by enry: https://github.com/src-d/enry/pull/205

bzz:
[enry v1.7.2](https://github.com/src-d/enry/releases/tag/v1.7.2) is out

PR #735, opened 2019-03-19
Add oniguruma bindings into generated gitbase binary

ajnavarro:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,7 @@ jobs:
       services: [docker]
 
       before_deploy:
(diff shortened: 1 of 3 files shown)


ajnavarro:
I'm having problems setting the version on the binary. The code appears to do the same but the version is not shown when `gitbase version` is called

jfontan:
Here's how @mcarmonaa solved it:

https://github.com/src-d/gitbase/pull/512

ajnavarro:
Problem fixed.

PR #737, opened 2019-03-19
dockerfile: include zero-config mysql client

### LINKED ISSUE #733: Feature request: include mysql-cli into docker image
From the user point of view, it makes sense to include mysql client in gitbase docker image so use would be able to connect/test and run queries without any additional dependencies or special configuration.
The official mysql image includes client: https://hub.docker.com/_/mysql

We are also going to use mysql-client inside engine, it would be nice to use the same gitbase image for that.

erizocosmico:
Fixes #733

With this change, users won't even have to install the MySQL client
on their machines or care about setting the parameters every single
time they use it. By just `docker exec -it $IMAGE_NAME mysql` it
will connect to gitbase without requiring any kind of configuration.

Connection details are stored in `$HOME/.my.cnf`, which is generated
at the start of the gitbase server. Because gitbase user and password
can change when you run it, not only when you build the image, this
needs to be done on startup. That's why `init.sh` has been included,
which just creates the configuration for MySQL client and starts the
gitbase server.


ajnavarro:
I'm concerned that this can be a security issue in the future... WDYT @smola ?

erizocosmico:
@ajnavarro you need access to the container to use this. If you have access to the container, there's bigger fish to fry that being able to query gitbase 😛 
How do you think this could be exploited?

smola:
Indeed. I don't see this as a security problem. `docker exec` access means root access in any case. You could just print the password or access the repositories volume directly.

See the equivalent with `postgres` container, for example:
```
$ docker run --name my-postgres --rm postgres:11-alpine
$ docker exec --user postgres -it my-postgres psql
```

smola:
Reminder: `docker run` access in practice means full root on the host. `docker exec` access means full root on any container, possibly also on the host.

PR #739, opened 2019-03-20
bblfsh: temporarily use aliases on some language names

### LINKED ISSUE #728: Parsing C# doesn't work
Run this query on any C# repository:
```
SELECT UAST(f.blob_content, LANGUAGE(f.file_path, f.blob_content)) AS uast
FROM refs AS r
NATURAL JOIN commit_files
NATURAL JOIN files AS f
WHERE r.ref_name = 'HEAD' AND f.file_path REGEXP('.*.cs')
LIMIT 5
```

it will return empty uasts.

erizocosmico:
Fixes #728

Until bblfsh lands support for aliases on all its methods, there
is no way of using certaing languages easily in gitbase, such as
C# or C++/C. By adding these aliases, it can be used until it's
fixed on bblfsh's side.

### DIFF
--- a/session.go
+++ b/session.go
@@ -96,9 +96,25 @@ type BblfshClient struct {
 	supportedLanguages []string
 }
 
+// FIXME: this is a temporary fix until bblfsh adds aliases to supported
+// languages. This should be removed once it is.
+var languageAliases = map[string]string{
+	"C#":    "csharp",
+	"JSX":   "JavaScript",
+	"C++":   "cpp",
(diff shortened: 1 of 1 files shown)


smacker:
maybe it's better to implement this workaround on bblfsh side? it would fix engine, gemini, jgit-spark-connector, gitbase-spark-connector. Not only gitbase.

kuba--:
I think it should be supported by enry.

PR #750, opened 2019-03-25
remove gitbase binary

kuba--:


smola:
You might want to add it to `.gitignore`.

PR #767, opened 2019-03-28
rule: parallelize projects calling UAST not under exchanges

### LINKED ISSUE #766: Parallelize expensive projects that are not under an exchange
```sql
SELECT uast_extract(
    uast(blob_content, 'csharp', "(//csharp:BinaryExpression_AddExpression/Left/uast:String | //csharp:InterpolatedStringExpression//csharp:InterpolatedStringTextToken[1])[starts-with(normalize-space(@Value), 'SELECT') or starts-with(normalize-space(@Value), 'select') or starts-with(normalize-space(@Value), 'UPDATE') or starts-with(normalize-space(@Value), 'update') or starts-with(normalize-space(@Value), 'DELETE') or starts-with(normalize-space(@Value), 'delete') or starts-with(normalize-space(@Value), 'INSERT') or starts-with(normalize-space(@Value), 'insert') or starts-with(normalize-space(@Value), 'CREATE') or starts-with(normalize-space(@Value), 'create') or starts-with(normalize-space(@Value), 'ALTER') or starts-with(normalize-space(@Value), 'alter') or starts-with(normalize-space(@Value), 'DROP') or starts-with(normalize-space(@Value), 'drop')]"),
    '@pos') AS positions,
    repository_id,
    file_path
FROM (
    SELECT f.repository_id,
        f.file_path,
        b.blob_content
    FROM (
        SELECT *
        FROM refs r
        NATURAL JOIN commit_blobs cb
        NATURAL JOIN blobs
        WHERE r.ref_name = 'HEAD'
            AND NOT IS_BINARY(blob_content)
    ) b
    INNER JOIN (
        SELECT repository_id, file_path, blob_hash
        FROM refs r
        NATURAL JOIN commit_files cf
        WHERE r.ref_name = 'HEAD'
    ) f
    ON b.blob_hash = f.blob_hash
        AND b.repository_id = f.repository_id
    WHERE language(f.file_path, b.blob_content) = 'C#'
) t
WHERE positions IS NOT NULL
```

This could be parallelized adding an exchange over the topmost projection. Instead, we do this serially, causing extremely low performance on queries using uast functions on the topmost projects under certain conditions.

erizocosmico:
Fixes #766

Since rows are iterated serially (except for exchange nodes), when
a Project node calls UAST, it's processed one at a time, which
underutilizes the resources in the machine.

When a Project calls UAST or UAST_MODE and is not under an Exchange
node, which already parallelizes its children, replaces the project
with a special node called parallelProject, which is essentially a
project that keeps up to N goroutines processing rows, where N is
the parallelism value of gitbase.

Signed-off-by: Miguel Molina <miguel@erizocosmi.co>

### DIFF
--- a/internal/rule/parallel_project.go
+++ b/internal/rule/parallel_project.go
@@ -1,0 +1,239 @@
+package rule
+
+import (
+	"context"
+	"io"
+	"strings"
+	"sync"
+
+	opentracing "github.com/opentracing/opentracing-go"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/plan"
+)
+
+type parallelProject struct {
+	*plan.Project
+	parallelism int
+}
+
+func newParallelProject(
+	projection []sql.Expression,
+	child sql.Node,
+	parallelism int,
+) *parallelProject {
+	return &parallelProject{
+		plan.NewProject(projection, child),
+		parallelism,
+	}
+}
+
+func (p *parallelProject) RowIter(ctx *sql.Context) (sql.RowIter, error) {
+	span, ctx := ctx.Span(
+		"plan.Project",
+		opentracing.Tag{
+			Key:   "projections",
+			Value: len(p.Projections),
+		},
+		opentracing.Tag{
+			Key:   "parallelism",
+			Value: p.parallelism,
+		},
+	)
+
+	iter, err := p.Child.RowIter(ctx)
+	if err != nil {
+		span.Finish()
+		return nil, err
+	}
+
+	return sql.NewSpanIter(
+		span,
+		newParallelIter(p.Projections, iter, ctx, p.parallelism),
+	), nil
+}
+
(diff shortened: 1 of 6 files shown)


internal/rule/parallel_project.go
```diff
@@ -0,0 +1,239 @@
...
+		errors:      make(chan error, parallelism),
+	}
+}
+
+func (i *parallelIter) Next() (sql.Row, error) {
+	if i.done {
+		return nil, io.EOF
+	}
+
+	if i.rows == nil {
```
kuba--: I think we need to wrap this `if` by mutex

erizocosmico: Nope, `Next` is never used in parallel

erizocosmico:
Should be passing now

ajnavarro:
Let's keep this on pause after have a new fixed RC

smola:
@erizocosmico Does the parallel project alter row order? If it does, it can give wrong results if ran after an ORDER BY.

erizocosmico:
@smola yes, you're right. It needs to check if it has an order by underneath and not execute it in parallel in that case or return the rows in the same order.

PR #773, opened 2019-04-03
do not build with oniguruma

erizocosmico:


Makefile
```diff
@@ -1,6 +1,7 @@
 # Package configuration
 PROJECT = gitbase
 COMMANDS = cmd/gitbase
+GO_BUILD_ARGS += -tags mysql_go_regex
```
ajnavarro: is this needed?

erizocosmico: Without this, all binaries except the static will have oniguruma

Issue #778, opened 2019-04-04
make static-package is not working properly


ajnavarro:
See: https://travis-ci.com/src-d/gitbase/jobs/190291212#L1566

juanjux:
The problem seems to be that https://hub.docker.com/r/srcd/gitbase/tags doesn't have that version.

ajnavarro:
Not fixed: https://travis-ci.com/src-d/gitbase/jobs/191104986#L1565

juanjux:
The `make ci-e2e` problem seems to be fixed, but not the one for `make static-package`. One question, since it seems `static-package` uses the docker hub image, where is that supposed to be created and pushed to dockerhub? (it's missing right now).

kuba--:
we don't push `-beta` (or with any other suffix) images.

PR #781, opened 2019-04-05
Upgrade mysql

kuba--:
This PR upgrades mysql dependency. New mysql-server contains the second part of closed iterators fix and makes go regex as a default engine.
At the same time (having go regex as a default) it closes https://github.com/src-d/gitbase/issues/778

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -672,7 +672,7 @@
   source = "github.com/src-d/go-git"
 
 [[projects]]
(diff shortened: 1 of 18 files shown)


juanjux:
Shouldn't this update the Makefile too to remove the go regex build tag?

kuba--:
yeah, I'll update it. but frankly this tag it's not used anymore

erizocosmico:
Version test is not passing

kuba--:
Yes, fixing (count is now int64)

kuba--:
fixed.

PR #782, opened 2019-04-08
Fix order of make commands

ajnavarro:
Attempt to fix https://github.com/src-d/gitbase/issues/778

Signed-off-by: Antonio Navarro Perez <antnavper@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,8 @@ jobs:
       services: [docker]
 
       before_deploy:
-        - make static-package
         - make docker-push-latest-release
+        - make static-package
 
       deploy:
         provider: releases


juanjux:
Mmm, if as you said on the issue comment docker push of beta releases is disabled (which is the case looking at `.ci/Makefile.am` this shouldn't do anything for `-beta1` should it?

If that's the case maybe the fix is to `make docker-build` so the image can be found locally before `make static-package` (but I can test locally because `make docker-build` is failing in my system with some error that I'm investigating).

ajnavarro:
Docker push on non final releases is not disabled, but we are not marking them as `latest`. Check `docker-push-latest-release` command:
```
docker-push-latest-release:
@DOCKER_PUSH_LATEST=$(IS_RELEASE) make docker-push
```

ajnavarro:
Check all versions here: https://hub.docker.com/r/srcd/gitbase/tags

Issue #788, opened 2019-04-09
Fix - remotes iterators


kuba--:
Test repos:
```
pga list -l ruby -f json | head -n 5 | jq -r '.sivaFilenames[]' | pga get -i -o repos
pga list -l python -f json | head -n 5 | jq -r '.sivaFilenames[]' | pga get -i -o repos
pga list -l javascript -f json | head -n 5 | jq -r '.sivaFilenames[]' | pga get -i -o repos
```

Test queries:
```
SELECT * FROM remotes WHERE remote_push_url REGEXP '^https://*';
SELECT * FROM remotes WHERE remote_push_url != 'trash';

CREATE INDEX uidx ON remotes USING pilosa (remote_push_url);
```

Both queries shall return 35 rows (based on test repos). After creating an index, the second query returns just 15 rows.

Note: 
The issue can be related with indexes (then we'll have to transfer it to mysql) or just with remotes table iterator.

PR #789, opened 2019-04-09
Don't skip repos for remotes table with more than 1 URL

kuba--:

Fixes https://github.com/src-d/gitbase/issues/788

For remotes with more than 1 URL, `remotesIndexKeyValueIter` skip repos.

### DIFF
--- a/remotes.go
+++ b/remotes.go
@@ -174,12 +174,12 @@ func (i *remotesRowIter) Close() error {
 }
 
 func remoteToRow(repoID string, config *config.RemoteConfig, pos int) sql.Row {
-	url := ""
+	var url interface{}
 	if pos < len(config.URLs) {
 		url = config.URLs[pos]
 	}
 
-	fetch := ""
+	var fetch interface{}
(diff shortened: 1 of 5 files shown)


kuba--:
@src-d/data-processing could you review my PR changes, again. I've added some fixes not only for index iterator (but also for `remotesRowIter`). It seems, we had a bug in _remotes iterators_.
PTAL.

remotes.go
```diff
@@ -177,13 +174,23 @@ func (i *remotesRowIter) Close() error {
 }
 
 func remoteToRow(repoID string, config *config.RemoteConfig, pos int) sql.Row {
+	url := ""
```
ajnavarro: If no URL, it should be `nil`, right?

kuba--: hmm, URLs are strings

ajnavarro: yep, I meant at SQL table level. If there is no value, we should return NULL instead of empty string.

kuba--: OK, done.

PR #795, opened 2019-04-11
 tools: make upgrade tool copy docs from go-mysql-server

### LINKED ISSUE #793: Function reference in gitbook documentation
Current _[Supported syntax](https://docs.sourced.tech/gitbase/using-gitbase/supported-syntax)_ links to [go-mysql-server documentation](https://github.com/src-d/go-mysql-server/blob/3dd441325d1731821eff0495fbf63747c258b8ff/SUPPORTED.md). It reads:

> To see the SQL subset currently supported take a look at this list from src-d/go-mysql-server.

I think we should include it right in the gitbook, and specially for the functions, have their signature and some description.

erizocosmico:
Fixes #793

There are some docs we want to have in gitbase as well as
go-mysql-server. To avoid possibly having different versions in
each project, this will automatically copy certain parts and files
from one to the other when there is an update.


erizocosmico:
Should be fixed now

erizocosmico:
Truly fixed now

erizocosmico:
Upgraded to latest go-mysql-server

PR #811, opened 2019-05-09
Diff prototype

kuba--:


kuba--:
WIP - tests missed, but ready to play with.

internal/function/diff.go
```diff
@@ -0,0 +1,292 @@
...
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+
+	"github.com/sergi/go-diff/diffmatchpatch"
+	"github.com/src-d/gitbase"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// statusMap is a lookup table for patch opertions
```
juanjux: s/opertions/operations

internal/function/diff.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	return from, to, nil
+}
+
+// diff compares two commits and returns a slice of diff items.
+// It compares changes (file by file) in parallel.
+// Another go routine collects diff items.
+// The function ignores binary files and equal operations.
+func diff(ctx context.Context, from, to *object.Commit) ([]*diffItem, error) {
```
mcuadros: Instead of writing this from scratch we should use [Commit.PatchContext](https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/object#Commit.PatchContext), in this case, you will find several problems to be fixed on `go-git` such as diff on binaries, or some other bugs, but this really worth it, since we are going to have many `diff` functions on gitbase

internal/function/diff.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	s, ok := ctx.Session.(*gitbase.Session)
+	if !ok {
+		return nil, gitbase.ErrInvalidGitbaseSession.New(ctx.Session)
+	}
+	return s.Pool.GetRepo(repoID)
+}
+
+func (d *Diff) getCommits(ctx *sql.Context, repo *gitbase.Repository, row sql.Row) (*object.Commit, *object.Commit, error) {
+	commitHash, err := exprToString(ctx, d.to, row)
```
mcuadros: The function should accept revisions and not only hashes, take a look at: [Repository.ResolveRevision](https://godoc.org/gopkg.in/src-d/go-git.v4#Repository.ResolveRevision).

This can be skipped in the first version if the usage of this is not straight forward.

internal/function/diff.go
```diff
@@ -0,0 +1,235 @@
...
+func (d *Diff) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
+	span, ctx := ctx.Span("gitbase.Diff")
+	defer span.Finish()
+
+	repo, err := d.getRepo(ctx, row)
+	if err != nil {
+		return nil, err
+	}
+
+	from, to, err := d.getCommits(ctx, repo, row)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internal/function/diff.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// IsNullable implements the Expression interface.
+func (*Diff) IsNullable() bool {
+	return false
+}
+
+// Resolved implements the Expression interface.
+func (d *Diff) Resolved() bool {
+	return d.to.Resolved() && d.to.Resolved() && (d.from == nil || d.from.Resolved())
```
golangcibot: SA4000: identical expressions on the left and right side of the '&&' operator (from `staticcheck`)

internal/function/diff.go
kuba--: @mcuadros - Ok, I've switched to `PatchContext` but I did it on Tree level, because on `object.Commit` level there is no check if `from` and/or `to` commits are `nil`, and `from` can be `nil` if there is no parent (the first commit).

internal/function/diff.go
kuba--: Added `ResolveRevision`, so first we'll try to resolve revision, if not the input string will be treated as a commit hash. But at the end the diff works with commit objects.

internal/function/diff.go
```diff
@@ -0,0 +1,262 @@
...
+		Status string `json:"status"`
+	}
+)
+
+// NewDiff constructor
+func NewDiff(args ...sql.Expression) (sql.Expression, error) {
+	d := &Diff{}
+	switch len(args) {
+	case 2:
+		d.repo, d.to = args[0], args[1]
```
mcuadros: Unless I am wrong, you can recover the repo name from `row[0]` as is done in the squash and works for me at: https://github.com/mcuadros/gitbase/commit/39ca34d520350c6089181591400d1056fa05c778#diff-eaff2b7fc045a1ea059d7a8d72e864ccR89

erizocosmico: That's only pure luck. You can never rely on the contents of the row to get the repository.

If you do this it won't work:

```sql
SELECT commit_stats(arguments) FROM (SELECT all args except repository id FROM tables) t
```

mcuadros: If can't be done in this way, we should provide the ability to recover the repositories without the requirement of passing the repository.

/cc @smola

mcuadros: If can't be done in this way, we should provide the ability to recover the repositories without the requirement of passing the repository.

/cc @smola

kuba--: Let's take a step back - `diff` is a function, so should work without any global context, independent. In other words diff(hash1, hash2) means nothing. You have to specify repo.
So why to take a repo from the different source than commits.
Either take all from args or from rows, but mix will be harder to test. 
Current implementation lets you just say `SELECT diff("myrepo", "hash1", "hash2")`
without repo it will require some data, under the hood. 
Last but not least, with select (without repo) like
`diff(hash1, hash2)` you don't know what was tested original repo, or maybe fork.

PR #835, opened 2019-05-27
functions: new COMMIT_STATS function

mcuadros:
New `COMMIT_STATS` function, it calculates the diff between commits based on a line comparisons algorithm instead of relying on the GNU diff one, being faster and more stable than using the [diff-match-patch](https://github.com/sergi/go-diff/tree/master/diffmatchpatch) integrated with go-git.

A line classification is also done using [gocloc](https://github.com/hhatto/gocloc) being almost free compared to the I/O cost.

The files of languages not supported by `gocloc` are included in the `Other` section and the vendor and binary files are just included in the `Files` count. 

This function contains similar code to https://github.com/src-d/gitbase/pull/811, so this will need to be refactored in an auxiliary algorithm, to avoid code duplication and input standardization. 

In terms of speed, traverse the history of gitbase computing the stats takes 22s compared to a version based on GNU diff of 44s. 

But this is not the only reason to do the analysis using the line based algorithm, the main reason is to calculate more granular stats by kind of line as shown here:

`SELECT COMMIT_STATS(repository_id, "HEAD") FROM repositories LIMIT 1;`

```js
{
  "Files": 5,
  "Code": {
    "Additions": 169,
    "Deletions": 0
  },
  "Comment": {
    "Additions": 5,
    "Deletions": 0
  },
  "Blank": {
    "Additions": 41,
    "Deletions": 0
  },
  "Other": {
    "Additions": 1,
    "Deletions": 0
  },
  "Total": {
    "Additions": 216,
    "Deletions": 0
  }
}
```


internal/utils/commit_stats_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"gopkg.in/src-d/go-git-fixtures.v3"
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/cache"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
+)
+
+func TestCalculateCommitStats(t *testing.T) {
+	fixtures.Init()
```
golangcibot: Error return value of `fixtures.Init` is not checked (from `errcheck`)

internal/utils/commit_stats_test.go
```diff
@@ -0,0 +1,122 @@
...
+			stats, err := CalculateCommitStats(r, from, to)
+			require.NoError(err)
+
+			assert.Equal(t, test.expected, stats)
+		})
+	}
+}
+
+func TestNewFileStats(t *testing.T) {
+	fixtures.Init()
```
golangcibot: Error return value of `fixtures.Init` is not checked (from `errcheck`)

internal/function/commit_stats_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}{
+		{
+			name: "init commit",
+			repo: expression.NewGetField(0, sql.Text, "repository_id", false),
+			from: nil,
+			to:   expression.NewGetField(1, sql.Text, "commit_hash", false),
+			row:  sql.NewRow("worktree", "b029517f6300c2da0f4b651b8642506cd6aaf45d"),
+			expected: &utils.CommitStats{
+				Files: 2,
+				Other: utils.KindStats{22, 0},
```
golangcibot: composites: `github.com/src-d/gitbase/internal/utils.KindStats` composite literal uses unkeyed fields (from `govet`)

internal/function/commit_stats_test.go
```diff
@@ -0,0 +1,67 @@
...
+				Total: utils.KindStats{22, 0},
```
golangcibot: composites: `github.com/src-d/gitbase/internal/utils.KindStats` composite literal uses unkeyed fields (from `govet`)

kuba--:
It's good that we have commit stats, but under the hood we still use the same `diff` library, so the solutions is _as slow as the prototype (https://github.com/src-d/gitbase/pull/811)_. Sometimes even slower by ~ 1s (because we calculate an extra stuff for `gocloc`.
For instance I tested against kubernetes with following query:
```sql
SELECT commit_author_name, commit_stats(repository_id, commit_hash)
FROM commits
WHERE repository_id="/private/tmp/repos/kubernetes"
LIMIT 10
```
with `LIMIT 10` it took `3.8s.` with `LIMIT 100` `35s.`

Also, I noticed that it cannot handle situations where you have a comment in the same line as code, e.g.:
```go
return nil, err // validation should avoid this
```
And it doesn't count it neither as code nor as comment. Check example query:
```sql
SELECT commit_stats("/private/tmp/repos/kubernetes","d87de1a90359291d1c9d7ff68cc927ddc9837f61") AS stats;
```
it says:
```json
{"Files":1,"Code":{"Additions":23,"Deletions":14},"Comment":{"Additions":0,"Deletions":0},"Blank":{"Additions":0,"Deletions":0},"Other":{"Additions":0,"Deletions":0},"Total":{"Additions":23,"Deletions":14}}
```
but in my opinion you have in code:
- Additions: 26
- Deletions: 17

Also it didn't count 1 Addition and 1 Deletion of blank lines.

mcuadros:
As I said, the goal is not being faster, is provide more information. Also, this avoids using diff-match-patch watch panics, in some cases.

We are using on the demo to build some graphs.

internal/utils/commit_stats.go
```diff
@@ -0,0 +1,366 @@
+package utils
```
erizocosmico: instead of a package `utils` which does not say much, what do you think about moving this to a `commitstats` package?

internal/utils/commit_stats.go
```diff
@@ -0,0 +1,366 @@
...
+	// Blank represents an empty line.
+	Blank
+	// Other represents a line from any other kind.
+	Other
+)
+
+// CalculateCommitStats calculates the CommitStats for from commit to another.
+// if from is nil the first parent is used, if the commit is orphan the stats
+// are compared against a empty commit.
+func CalculateCommitStats(r *git.Repository, from, to *object.Commit) (*CommitStats, error) {
```
erizocosmico: If it's moved to `commitstats` package, I'd rename this to `Calculate` to avoid redundancy in `commitstats.CalculateCommitStats`.

internal/function/commit_stats.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/src-d/gitbase"
+	"github.com/src-d/gitbase/internal/utils"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+type CommitStats struct {
```
erizocosmico: Missing docs on the type, constructor and a few methods

internal/function/commit_stats.go
```diff
@@ -0,0 +1,144 @@
...
+	repo sql.Expression
```
erizocosmico: Expression properties of UDFs should all be public, so they can be accessed during transforms

internal/utils/commit_stats.go
mcuadros: inside of utils package?

erizocosmico: No, directly in internals

mcuadros: More packages are about to come, will be a mess having many packages that the relation is being utils to the others.

erizocosmico: We're not gonna have that many packages. That's the very definition of internal: a collection of packages that are only meant for internal use by the package. Having an `utils` package does not say much other than the fact that the package is useful for some other package, which is why the package exists.

ajnavarro:
@mcuadros could you have a look into linter and people messages? This is now a high priority.

internal/function/commit_stats_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}{
+		{
+			name: "init commit",
+			repo: expression.NewGetField(0, sql.Text, "repository_id", false),
+			from: nil,
+			to:   expression.NewGetField(1, sql.Text, "commit_hash", false),
+			row:  sql.NewRow("worktree", "b029517f6300c2da0f4b651b8642506cd6aaf45d"),
+			expected: &commitstats.CommitStats{
+				Files: 2,
+				Other: commitstats.KindStats{22, 0},
```
golangcibot: composites: `github.com/src-d/gitbase/internal/commitstats.KindStats` composite literal uses unkeyed fields (from `govet`)

internal/function/commit_stats_test.go
```diff
@@ -0,0 +1,67 @@
...
+				Total: commitstats.KindStats{22, 0},
```
golangcibot: composites: `github.com/src-d/gitbase/internal/commitstats.KindStats` composite literal uses unkeyed fields (from `govet`)

kuba--:
Btw. regarding incorrect file stats (numbers) - I've sent PR with fix: https://github.com/hhatto/gocloc/pull/19

kuba--:
Actually `gocloc` uses `enry`, so internally we have information about language (but we loose it)

mcuadros:
Sorry @ajnavarro what documentation is missing? Also, this function is not using diff.
About the requirements of doing it by language are about to come, once that this is merge.

ajnavarro:
With missing documentation, I mean https://github.com/src-d/gitbase/blob/master/docs/using-gitbase/functions.md and add some examples.

kuba--:
It's not using diff?
How about following lines
https://github.com/src-d/gitbase/pull/835/files#diff-5073f9b8c2c86946584dc39d2c0610c3R75

https://github.com/src-d/gitbase/pull/835/files#diff-5073f9b8c2c86946584dc39d2c0610c3R106

mcuadros:
Its using `diff_tree` but not `diff`

smola:
By the way, the confusion on diff usage is my bad, since I mentioned it this morning.

PR #848, opened 2019-06-02
Upgrade gocloc

kuba--:

 New gocloc contains a fix for nested comments and for comments in the same line as code.


erizocosmico:
Just take into account that this will need to be fixed again after the PR to update to modules is merged.

kuba--:
yes, I hope we'll switch to go modules asap. so we'll be able to upgrade enry to 2.0.0, gloc to v0.3
and new bblfsh.

PR #863, opened 2019-06-13
internal/function: take into account if repository is resolved

erizocosmico:
Fixes #861 
 
 - [x] I updated the documentation explaining the new behavior if any.
 - [x] I updated CHANGELOG.md file adding the new feature or bug fix.
 - [x] I updated go-mysql-server using `make upgrade` command if applicable.
 - [x] I added or updated examples if applicable.
 - [x] I checked that changes on schema are reflected into the documentation, if applicable.

PR #864, opened 2019-06-13
improve commit stats

### LINKED ISSUE #859: Document COMMIT_STATS
Add missing documentation about COMMIT_STATS

### LINKED ISSUE #860: Improve error handling on COMMIT_STATS
now, any error on that function makes the query crash. Handle better errors and return nil column if stats cannot be processed.

erizocosmico:
Fixes #859 
Fixes #860 

Depends on: #863 

 
 
### DIFF
--- a/docs/using-gitbase/functions.md
+++ b/docs/using-gitbase/functions.md
@@ -6,6 +6,7 @@ To make some common tasks easier for the user, there are some functions to inter
 
 |     Name     |                                               Description                                                                      |
 |:-------------|:-------------------------------------------------------------------------------------------------------------------------------|
+|`commit_stats(repository_id, [from_commit_hash], to_commit_hash)`|returns the stats between two commits for a repository. If from is empty, it will compare the given `to_commit_hash` with its parent commit|
 |`is_remote(reference_name)bool`| check if the given reference name is from a remote one                                                          |
 |`is_tag(reference_name)bool`| check if the given reference name is a tag                                                                         |
 |`is_vendor(file_path)bool`| check if the given file name is a vendored file                                                                         |
--- a/internal/function/commit_stats.go
+++ b/internal/function/commit_stats.go
@@ -3,6 +3,7 @@ package function
 import (
 	"fmt"
 
+	"github.com/sirupsen/logrus"
 	"github.com/src-d/gitbase"
 	"github.com/src-d/gitbase/internal/commitstats"
 
@@ -81,7 +82,7 @@ func (f *CommitStats) Children() []sql.Expression {
 
 // IsNullable implements the Expression interface.
 func (*CommitStats) IsNullable() bool {
-	return false
+	return true
 }
 
 // Resolved implements the Expression interface.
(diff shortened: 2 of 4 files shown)


docs/using-gitbase/functions.md
```diff
@@ -6,6 +6,7 @@ To make some common tasks easier for the user, there are some functions to inter
 
 |     Name     |                                               Description                                                                      |
 |:-------------|:-------------------------------------------------------------------------------------------------------------------------------|
+|`commit_stats(repository_id, [from_commit_hash], to_commit_hash)`|returns the stats between two commits for a repository. If from is empty, it will compare HEAD with the given `to_commit_hash`|
```
kuba--: If `from` is missed then `to` is compared with a parent:
https://github.com/src-d/gitbase/blob/494e149875f5ea4bfe012051aeb59b662d1e64cd/internal/commitstats/commit_stats.go#L40

internal/function/commit_stats.go
```diff
@@ -96,20 +99,39 @@ func (f *CommitStats) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 
 	r, err := f.resolveRepo(ctx, row)
 	if err != nil {
-		return nil, err
+		ctx.Warn(0, "commit_stats: unable to resolve repository")
```
kuba--: Lets wrap this by `logWarn(ctx, ...)` or sth. like this

erizocosmico: Just to avoid the `0, `?
In this case the messages are the same, in the rest of the errors it's not. Add the function just for this case doesn't seem worth it imho

docs/using-gitbase/functions.md
erizocosmico: ooopsie

internal/function/commit_stats.go
kuba--: I just thought to have a function for logs and session warnings

erizocosmico: yup, the thing is the log and the warnings are not the same messages

if you have

ctx.Warn(0, "something", with, args)
logrus.WithFields(some fields).Error("different message")

you will need something like:

logWarn(ctx, "message for warn", []interface{}{args, for, warn}, "message for log", logrus.Fields{args, for, log})

And it kinda defeats the purpose

internal/function/commit_stats.go
```diff
@@ -96,20 +99,39 @@ func (f *CommitStats) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
...
 	from, err := f.resolveCommit(ctx, r, row, f.From)
 	if err != nil {
-		return nil, err
+		ctx.Warn(0, "commit_stats: unable to resolve 'from' commit of repository: %v", r)
+		log.WithField("err", err).Error("commit_stats: unable to resolve from commit")
+		return nil, nil
 	}
 
-	return commitstats.Calculate(r.Repository, from, to)
+	result, err := commitstats.Calculate(r.Repository, from, to)
```
ajnavarro: This will fail for root commits too, when commit.Parent(0) fails.

erizocosmico: Yep, that's why the error is not returned

erizocosmico:
rebased

PR #867, opened 2019-06-14
[go-borges] cli: added support to specify format per directory

### LINKED ISSUE #866: [go-borges] Change arguments to specify libraries
To pass directories with repos `--directory` will be used. By default it will add non bare git repositories but this can be changed with:

```
--format (siva|git)
--bare false
--rooted true
--bucket 2
```

* `git` format can specify if they are bare or not.
* `siva` format cam specify bucketing level and if it's rooted or not

Defaults are the ones in the example. Next version will use auto discovery of bareness.

jfontan:
Global format can be specified with `--format`, `--bare`, `--bucket` and `--rooted`. To mix directories the format can be specified as a file URL:

    -d file:///path/to/library?format=siva&bucket=4&rooted=false

Also:

* Updated builder container to go 1.12
* Added SIVA env to docker init.sh to configure gitbase to use siva library
 
Fixes #866


docs/using-gitbase/configuration.md
```diff
@@ -86,37 +86,35 @@ Help Options:
...
-          --depth=                                     load repositories looking at less than
-                                                       <depth> nested subdirectories. (default: 1000)
-          --host=                                      Host where the server is going to listen
-                                                       (default: localhost)
-      -p, --port=                                      Port where the server is going to listen
-                                                       (default: 3306)
+      -d, --directories=                               Path where standard git repositories are located,
+                                                       multiple directories can be defined.
+          --format=[git|siva]                          Library format (default: git)
+          --bucket=                                    Bucketing level to use with siva libraries (default: 2)
```
kuba--: do we really need it? If yes, I would suggest to have at least 2 (so you can't set up less than 2).

docs/using-gitbase/getting-started.md
```diff
@@ -116,3 +116,20 @@ As of MySQL 8.0 [the default authentication method is `caching_sha2_password`](h
...
+## Library format specification
+
+By default the directories added to github should contain normal git repositories. If the format of the repositories is different you have two ways to specify it.
+
+If all the directories are in the same format you can set it globally with these parameters:
+
+* `--format`: it can be either `git` for filesystem repositories or `siva` for siva archives
+* `--bare`: specifies that git archives are bare, can only be used with `git` format
+* `--bucket`: sets the number of characters to use for bucketing, used with `siva` libraries
+* `--rooted`: enables or disables rooted reposotories management in `siva` libraries
```
kuba--: typo `repositories`

cmd/gitbase/command/server.go
```diff
@@ -284,26 +312,116 @@ func (c *Server) addDirectories() error {
...
+	query := u.Query()
+
+	for k, v := range query {
+		if len(v) != 1 {
+			logrus.Errorf("invalid number of options for %v", v)
+			return dir, ErrInvalid
+		}
+
+		val := v[0]
+		switch k {
```
kuba--: `strings.ToLower` - WDYT?

docs/using-gitbase/configuration.md
jfontan: I don't think we should impose any limits. We are doing this so we can also read old sivas, for example from PGA. A use case can be copying some siva files in a directory without bucketing so `bucket=0` should be available. The default parameters for siva are compatible with the new download.

kuba--: but it's the same if I say `bucket=1234567879`

jfontan: I understand adding a maximum and probably not letting negative numbers that can crash the program. But I don't find useful setting the minimum bucketing level to 2.

PR #870, opened 2019-06-17
Upgrade enry to v2

kuba--:
`gocloc` already uses `enry/v2`, so upgrading enry will let us use only one enry version across gitbase and dependencies (gocloc).


ajnavarro:
@kuba-- could you rebase please?

Issue #815, opened 2019-05-16
Expose metrics on cache hit/miss rates


smola:
We have quite a few caches in gitbase, but as far as I can tell, we have no way to get metrics for cache hit/miss rates. Those would be very useful for troubleshooting and tuning.

kuba--:
where do you want to store those metrics?
gitbase uses only open tracing, there is no statsD/graphana infra (as far as I know).
We can push this kind of metrics to StatsD or any other time series DB over UDP.

smola:
In our clusters we use Prometheus, which can be setup to retrieve metrics from an HTTP endpoint.

kuba--:
sure thing, my question was- do we want to expose http port?

smola:
@kuba-- Yes, I think so, since we're going that route for other components.

PR #877, opened 2019-06-18
Expose (if enabled) prometheus metrics over http (#815).

kuba--:

Closes https://github.com/src-d/gitbase/issues/815
 
 - [X] I updated the documentation explaining the new behavior if any.
 - [X] I updated go-mysql-server using `make upgrade` command if applicable.
 - [X] I added or updated examples if applicable.
 - [X] I checked that changes on schema are reflected into the documentation, if applicable.


internal/function/uast.go
```diff
@@ -263,6 +289,8 @@ func (u *uastFunc) getUAST(
 		return nil, nil
 	}
 
+	finish(ok)
```
erizocosmico: what happens if there is an early return?

kuba--: early return happens on error, and here we build uast histogram (measure latency), so if we collect immediately returns on errors then final histogram  may be incorrect. That's why we measure only successfully queries, here.
If we want, we can count errors, but I think it should be done on uast side, but it's up to us.

ajnavarro:
LGTM after rebase @kuba--

ajnavarro:
@kuba-- could you rebase please?

kuba--:
PTAL a new PR: https://github.com/src-d/gitbase/pull/919

PR #893, opened 2019-06-20
cli: log repositories found on startup

jfontan:
Uses the same log message as before except that instead of `path` it's `id`.
 
### DIFF
--- a/cmd/gitbase/command/server.go
+++ b/cmd/gitbase/command/server.go
@@ -290,7 +290,16 @@ func (c *Server) addDirectories() error {
 		}
 	}
 
-	return nil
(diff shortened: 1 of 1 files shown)


cmd/gitbase/command/server.go
```diff
@@ -290,7 +290,16 @@ func (c *Server) addDirectories() error {
 		}
 	}
 
-	return nil
+	repos, err := c.rootLibrary.Repositories(borges.ReadOnlyMode)
+	if err != nil {
+		return err
+	}
+
```
kuba--: `defer repos.Close()`?

jfontan: You're right! Added.

PR #897, opened 2019-06-21
cli: detect bareness of git libraries

jfontan:
Now by default the type of git repositories (bare or non bare) is chosen automatically using the first repository found in the directory. To force an specific format `--bare` and `--non-bare` can be used. Also the URL format now has 3 possible values for `bare` option: `true`, `false` and `auto`.

```
$ gitbase server --non-bare -d //dir/1
$ gitbase server -d 'file:///dir/1?bare=auto' -d 'file:///dir/1?bare=true'
 ```

### CODE REVIEW

File: cmd/gitbase/command/server_test.go [hunk 1]
```diff
@@ -1,9 +1,14 @@
...
 
 import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	fixtures "gopkg.in/src-d/go-git-fixtures.v3"
```
erizocosmico: fixtures should not be v3, but "github.com/src-d/go-git-fixtures", which is the one we're using throughout the code

jfontan: Right! I should stop relying blindly in `goimports`. Fixed.

cmd/gitbase/command/server.go
```diff
@@ -344,12 +362,20 @@ func (c *Server) addDirectory(d directory) error {
 	return nil
 }
 
+type bareness int
```
kuba--: hmm, maybe just `bareOpt`

cmd/gitbase/command/server.go
```diff
@@ -395,12 +421,19 @@ func parseDirectory(dir directory) (directory, error) {
 			dir.Format = val
 
 		case "bare":
-			if val != "true" && val != "false" {
+			if val != "true" && val != "false" && val != "auto" {
```
kuba--: why do we need this `if`, if we can add a `default` to the `switch`?

cmd/gitbase/command/server.go
```diff
@@ -59,6 +59,7 @@ type Server struct {
 	Format        string         `long:"format" default:"git" choice:"git" choice:"siva" description:"Library format"`
 	Bucket        int            `long:"bucket" default:"2" description:"Bucketing level to use with siva libraries"`
 	Bare          bool           `long:"bare" description:"Sets the library to use bare git repositories, used only with git format libraries"`
+	NonBare       bool           `long:"non-bare" description:"Sets the library to use non bare git repositories, used only with git format libraries"`
```
mcarmonaa: Why is this necessary? Bare == true doesn't mean NonBare == false?

jfontan: By default it autodetects the type (bare or non bare). To be able to know if we explicitly set one of the options we can either use two booleans or an argument with an option `--bare=true`. Using booleans seems more similar to the other options.

cmd/gitbase/command/server.go
jfontan: I found `bareness` funny but `bareOpt` may be clearer. :+1:

cmd/gitbase/command/server.go
jfontan: Totally right. I modified the previous option that only had two possibilities. :+1:

jfontan: Changed

cmd/gitbase/command/server.go
jfontan: Changed

jfontan:
I've rebased with latest `master`

PR #898, opened 2019-06-21
gitbase: use only one cache for all repositories

jfontan:
Also fix a bug getting the cache size from cli.

Before integrating the use of `go-borges` repositories were managed by `RepositoryPool`. When it was created it initialized a git object cache that is used by all the repositories. This cache is also used when using indexes and reading directly from packfiles (cache is retrieved with `Repository.Cache()`, https://github.com/src-d/gitbase/blob/master/packfiles.go#L239).

Now `go-borges` takes care of repository libraries and retrieving go-git repositories from them. To make it work the same we initialize `go-borges` library with a cache generated in `gitbase` (`command.Server`) but I forgot that `RepositoryPool` was also generating a cache internally and returning it when calling `Repository.Cache()`. This means that there was two different objects caches, one for repositories opened without indexes and other for reading packfiles when the index was used. The problems are:

* It could theoretically use twice the amount of memory
* The objects in one of the cache won't be seen by the other so they will be read twice from disk

The change makes `RepositoryPool` get an initialized cache that will be used by all repos (both with and without index).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - git libraries bare or non bare format is automatically detected ([#897](https://github.com/src-d/gitbase/pull/897))
+- Fix bug that created multiple object cache with incorrect size ([#898](https://github.com/src-d/gitbase/pull/898))
 
 ## [0.22.0-beta1] - 2019-06-20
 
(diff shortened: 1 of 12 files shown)


juanjux:
One question (not doubting, just want to understand better): what's the rationale for this change?

jfontan:
> One question (not doubting, just want to understand better): what's the rationale for this change?

You are right, there was almost no description to the PR :see_no_evil:. I've updated the description with the problem it's fixing.

juanjux:
@jfontan awesome description, thanks!

jfontan:
I've just rebased with latest `master`.

PR #914, opened 2019-07-03
gitbase: report repository id on errors

### LINKED ISSUE #909: Report repository_id on errors
When a query fails because of a git error, like the infamous `object not found` or possibly other deeper I/O errors, it would be very helpful for troubleshooting if repository_id (or path or any other proxy) was added to the error.

erizocosmico:
Closes #909 

I started implementing this in another way, just wrapping error by error with the error to just do it on git errors, but it was a ton of changes and very prone to not be done in the future.

This way, the iterator is wrapped and the errores are always added the info, no matter how much we change the iterators and the errors they return.


partition.go
```diff
@@ -271,3 +271,36 @@ func (i *partitionedIndexKeyValueIter) Close() error {
...
+	iter sql.RowIter
+	repo *Repository
+}
+
+func newRepoRowIter(repo *Repository, iter sql.RowIter) sql.RowIter {
+	return &repoRowIter{iter, repo}
+}
+
+func (i *repoRowIter) Next() (sql.Row, error) {
+	row, err := i.Next()
```
golangcibot: SA5007: infinite recursive call (from `staticcheck`)

kuba--:
LGTM - what do you think about renaming to sth. like `errorWithRepo(...)`

erizocosmico:
Done

PR #916, opened 2019-07-03
Varchar

### LINKED ISSUE #824: Better types for table columns (VARCHAR instead of TEXT)
Currently query
```
describe table commits;
```
produces

```
+---------------------+----------+
| name                | type     |
+---------------------+----------+
| repository_id       | TEXT     |
| commit_hash         | TEXT     |
| commit_author_name  | TEXT     |
| commit_author_email | TEXT     |
| commit_author_when  | DATETIME |
| committer_name      | TEXT     |
| committer_email     | TEXT     |
| committer_when      | DATETIME |
| commit_message      | TEXT     |
| tree_hash           | TEXT     |
| commit_parents      | JSON     |
+---------------------+----------+
```

Which is technically correct but columns like `repository_id`, `commit_hash`, `commit_author_name`, ... better be `VARCHAR`

By semantic TEXT mean huge chunk of text and due to it unique. While `VARCHAR` holds reasonable amount of text and often repeats.
No wonder people rely on it. For example Superset [makes columns gropable](https://github.com/src-d/superset-compose/blob/master/superset/superset/connectors/sqla/models.py#L879) and filterable when they are `VARCHAR` but doesn't do it for `TEXT`.

Source{d} product has a [bug](https://github.com/src-d/superset-compose/issues/34) because of it. And it might be useful in some other use cases too.

juanjux:
Fixes #824.

Switch some types to `VARCHAR` with a size, mostly those of know length or maximum length like hashes (40), entry_modes (16), and emails (254 according to the last RFC). Probably some types like paths, repository_ids and URLS could be switched too, but we should choose a "safe" length for those. One option is to check the length used in git source code; I did that for a couple like the git message length but it used `size_t`.
 
 - [X] I updated the documentation explaining the new behavior if any.
 - [X] I updated go-mysql-server using `make upgrade` command if applicable.
 - [X] I checked that changes on schema are reflected into the documentation, if applicable.


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Now non rooted siva files support old siva rooted repositories.
+- Switch some types of know length or maximum length (mostly hashes and emails)
```
erizocosmico: s/know/known/

juanjux: Fixed, thanks.

ajnavarro:
@juanjux could you check tests please?

juanjux:
@ajnavarro I don't know wtf is happening with the Travis build and the `TestBuildSquashedTable` test because, of course, It Works On My Machine :tm: but I'm investigating.

juanjux:
Could reproduce the problem by going to the specific directory of the test. Now I just need to understand wtf those numbers differ.

erizocosmico:
Can you rebase?

juanjux:
Ok, it failed again, daaaamn

juanjux:
Worked, @ajnavarro can I have a sweet merge?

ajnavarro:
The schema and docu changed too! well done!

PR #920, opened 2019-07-05
fix files table by removing git log usage

erizocosmico:
This may have been what was causing the test failures in gsc.
 

erizocosmico:
can confirm it solves the issue on gsc

kuba--:
so there is a bug in `git.LogOptions{All: true,}` ?

erizocosmico:
@kuba-- not really, but we changed the behaviour of commit traversing a while ago and we didn't make the change here. Also, it only happens with new borges and old siva files. We have no idea why, but this solves it 🤷‍♂

Issue #930, opened 2019-07-11
Use struct tags on commit_stats and commit_file_stats


erizocosmico:
Right now all JSON produced by these UDFs has the fields in Uppercase, which is weird when using JSON (specially in Spark, using the struct fields). Should we add some struct tags so it's in `snake_lower_case`?

WDYT @src-d/data-processing?

ajnavarro:
As I answered into the PR, I'm concerned about backward compatibility. All queries that are using these functions will stop working with no error message.

PR #931, opened 2019-07-11
internal/function: use custom fixed struct as LOC result

### LINKED ISSUE #927: LOC function is returning external library object that might change
This query from the documentation:
```
SELECT
    LANGUAGE(file_path, blob_content) as lang,
    SUM(JSON_EXTRACT(LOC(file_path, blob_content), '$.Code')) as code,
    SUM(JSON_EXTRACT(LOC(file_path, blob_content), '$.Comments')) as comments,
    SUM(JSON_EXTRACT(LOC(file_path, blob_content), '$.Blanks')) as blanks,
    COUNT(1) as files
FROM commit_files
NATURAL JOIN refs
NATURAL JOIN blobs
WHERE ref_name='HEAD'
GROUP BY lang;
```
Actually is not working.

After updating goloc (a library that is used under the hood), they changed the output struct and instead of `Code` the output now is `code` and so on.

We should wrap that output struct to be sure that output JSON won't change if library changes.

erizocosmico:
Fixes #927 
 
### DIFF
--- a/docs/using-gitbase/functions.md
+++ b/docs/using-gitbase/functions.md
@@ -169,6 +169,26 @@ Also, if you want to retrieve values from a non common property, you can pass it
 
 > uast_extract(nodes_column, 'some-property')
 
+## How to use `loc`
+
+`loc` will return statistics about the lines of code in a file, such as the code lines, comment lines, etc.
+
+It requires a file path and a file content.
+
+> loc(file_path, blob_content)
+
+The result of this function is a JSON document with the following shape:
+
+```
+{
+	"Code": code lines,
+	"Comment": comment lines,
+	"Blank": blank lines,
+	"Name": file name,
+	"Lang": language
+}
+```
+
 ## How to use `commit_file_stats`
 
 `commit_file_stats` will return statistics about the line changes in all files in the given range of commits classifying them in 4 categories: code, comments, blank lines and other.
--- a/internal/function/loc.go
+++ b/internal/function/loc.go
@@ -52,6 +52,15 @@ func (f *LOC) WithChildren(children ...sql.Expression) (sql.Expression, error) {
 	return NewLOC(children...)
 }
 
+// LocFile is the result of the LOC function for each file.
+type LocFile struct {
+	Code     int32  `json:"Code"`
+	Comments int32  `json:"Comment"`
+	Blanks   int32  `json:"Blank"`
+	Name     string `json:"Name"`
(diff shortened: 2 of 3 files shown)


docs/using-gitbase/examples.md
```diff
@@ -81,9 +81,9 @@ GROUP BY committer_email,
 ```sql
 SELECT
     LANGUAGE(file_path, blob_content) as lang,
-    SUM(JSON_EXTRACT(LOC(file_path, blob_content), '$.Code')) as code,
```
ajnavarro: I would say to do the opposite, keep output as it was. Also, other functions have keys with first letter uppercase too.

erizocosmico: Using uppercase is a bit weird in JSON, hence https://github.com/src-d/gitbase/issues/930

If we do want to use uppercase, though, I can close the issue and change it back here.

ajnavarro: I'm concerned about backward compatibility. All queries that are using these functions will stop working with no error message...

erizocosmico:
Struct tags are now uppercase. Docs added with the shape of LOC.

docs/using-gitbase/examples.md
```diff
@@ -81,9 +81,9 @@ GROUP BY committer_email,
...
-    SUM(JSON_EXTRACT(LOC(file_path, blob_content), '$.Comments')) as comments,
```
ajnavarro: Now you need to keep this example as it was.

docs/using-gitbase/functions.md
```diff
@@ -17,7 +17,7 @@ To make some common tasks easier for the user, there are some functions to inter
 |`uast_xpath(blob, xpath) blob`| performs an XPath query over the given UAST nodes                                                                |
 |`uast_extract(blob, key) text array`| extracts information identified by the given key from the uast nodes                                       |
 |`uast_children(blob) blob`| returns a flattened array of the children UAST nodes from each one of the UAST nodes in the given array              |
-|`loc(path, blob) json`| returns a JSON map, containing the lines of code of a file, separated in three categories: Code, Blank and Comment lines |
+|`loc(path, blob) json`| returns a JSON map, containing the lines of code of a file, separated in three categories: code, blank and comment lines |
```
ajnavarro: and this.

erizocosmico:
Done

PR #935, opened 2019-07-19
Let the user define LOG_LEVEL=info using docker

### LINKED ISSUE #934: Can not use `info` logging level from docker image
caused by https://github.com/src-d/sourced-ce/issues/142

Under the context UX on **source{d} CE**, I'm reviewing its logs (to make them as simple as possible, to be understandable for the user), and I saw that for `gitbase`, is not possible to use debug level `info` (`info`, which is the default for `gitbase`, is the only one failing, the others are not)

I found that it's because the `verbose` mode is enabled in the [`init.sh`](https://github.com/src-d/gitbase/blob/master/init.sh#L13), so when using `gitbase` docker image, the `debug` level is set from the beginning by [server.go:131](https://github.com/src-d/gitbase/blob/master/cmd/gitbase/command/server.go#L131), and since it was passed `info` to the command (the default), it is not recovered because of the `if` at [server.go:140](https://github.com/src-d/gitbase/blob/master/cmd/gitbase/command/server.go#L140)

If it is requested any other log level, the `if` at [server.go:140](https://github.com/src-d/gitbase/blob/master/cmd/gitbase/command/server.go#L140) will let that the desired log level is used.

dpordomingo:
fix #934

- docker entrypoint should not force the usage of 'verbose' mode
- docker entrypoint uses 'debug' log level as default if not defined (for backward compatibility)
- verbose flag will override any other passed log level, not only 'info'

 TODO;
 
n/a
 - I added or updated examples if applicable.
 - I checked that changes on schema are reflected into the documentation, if applicable.
- I updated go-mysql-server using `make upgrade` command if applicable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- When using `-v` verbose flag, will set `debug` as logging level, ignoring any other passed.
+
+### Fixed
+
+- If using docker image, and `info` logging level, it will be now used instead of `debug`.
+
+
 ## [0.24.0-beta1] - 2019-07-08
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- When using `-v` verbose flag, will set `debug` as logging level, ignoring any other passed.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- When using `-v` verbose flag, will set `debug` as logging level, ignoring any other passed.
+
+### Fixed
+
+- If using docker image, and `info` logging level, it will be now used instead of `debug`.
```
smola: Can you add the reference to the issue here?
Lately we have a lot of release notes without any issue reference and it makes it quite hard to expand info about each item.

dpordomingo: True! I always did it in Landing, and forgot it here.
Done by 52f548e

PR #944, opened 2019-08-02
function: implement uast_imports UDF

### LINKED ISSUE #943: UDF to extract imports from an UAST
Followup of https://github.com/src-d/gitbase/issues/933

erizocosmico:
Closes #943

After changes to the structure of imports, the way we used to collect
them is no longer valid, so now we're using the uast.AllImportPaths
helper included in the bblfsh sdk to obtain import paths from nodes.
 

juanjux:
Travis failing with some connection problem, retry would probably fix it.

PR #949, opened 2019-08-14
Fix documentation typos and make style more consistent

agarciamontoro:
I've been reading gitbase documentation to get used to the project and the codebase, so I fixed the typos and errors I found.

Changes:

- Fix a malformed query in examples.md. It used the non-existing `commit_blob` column where it should use `blob_hash`.
- Make description of functions consistent throughout all the tables in functions.md. I decided to stick to the lower-case present-tense style, as each row is easily read as `Function X returns Y`. This is opinionated (consistency should be not), so I'm ok discussing the best style and changing it again once we agree on one.
- Fix several typos.
 
(Not adding the PR template here because this only updates typos in documentation, so none of the items apply here)

### DIFF
--- a/docs/using-gitbase/functions.md
+++ b/docs/using-gitbase/functions.md
@@ -6,19 +6,19 @@ To make some common tasks easier for the user, there are some functions to inter
 
 |     Name     |                                               Description                                                                      |
 |:-------------|:-------------------------------------------------------------------------------------------------------------------------------|
-|`commit_stats(repository_id, [from_commit_hash], to_commit_hash) json`|returns the stats between two commits for a repository. If from is empty, it will compare the given `to_commit_hash` with its parent commit. Vendored files stats are not included in the result of this function. This function is more thoroughly explained later in this document.|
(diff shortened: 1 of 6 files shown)


agarciamontoro:
Oops, forgot to sign-off the commit. Solved.

docs/using-gitbase/functions.md
```diff
@@ -6,90 +6,90 @@ To make some common tasks easier for the user, there are some functions to inter
...
+|`loc(path, blob) json`| returns a JSON map, containing the lines of code of a file, separated in three categories: Code, Blank and Comment lines. |
+|`version() text`| returns the gitbase version in the following format `8.0.11-{GITBASE_VERSION}` for compatibility with MySQL versioning. |
 ## Standard functions
 
 These are all functions that are available because they are implemented in `go-mysql-server`, used by gitbase.
 
 <!-- BEGIN FUNCTIONS -->
 |     Name     |                                               Description                                                                      |
 |:-------------|:-------------------------------------------------------------------------------------------------------------------------------|
-|`ARRAY_LENGTH(json)`|If the json representation is an array, this function returns its size.|
```
ajnavarro: this code is auto copied from go-mysql-server. You should fix that there and this will be applied here after updating the dependency.

agarciamontoro: Oh, ok! I'll move this change there, thanks!

agarciamontoro:
Removed the changes in standard functions in functions.md and all changes in supported_clients.md here. Opened https://github.com/src-d/go-mysql-server/pull/804.

PR #968, opened 2019-10-09
Don't abort on git.Blame errors

### LINKED ISSUE #966: Blame: error kills connection and stop query.
Executing the following query:

`select commit_hash,explode(blame(repository_id, commit_hash)) from commits limit 100000`

I got this error:

```
INFO[0003] audit trail                                   action=authentication address="127.0.0.1:60820" success=true system=audit user=root
DEBU[0003] executing query                               query="select commit_hash,explode(blame(repository_id, commit_hash)) from commits limit 100000"
INFO[0003] audit trail                                   action=authorization address="127.0.0.1:60820" connection_id=1 permission=read pid=1 query="select commit_hash,explode(blame(repository_id, commit_hash)) from commits limit 100000" success=true system=audit user=root
INFO[0007] audit trail                                   action=query address="127.0.0.1:60820" connection_id=1 duration=3.690559548s err="contents and commits have different length" pid=1 query="select commit_hash,explode(blame(repository_id, commit_hash)) from commits limit 100000" success=false system=audit user=root
ERRO[0007] Error in the middle of a stream to client 1 (127.0.0.1:60820): contents and commits have different length 
INFO[0007] kill query: pid 1                            
INFO[0007] ConnectionClosed: client 1              
```
This error shouldn't stop te query, just add a Warning to check later on SHOW WARNINGS.

juanjux:
Fixes #966 

Signed-off-by: Juanjo Alvarez <juanjo@sourced.tech>

 
 
### DIFF
--- a/internal/function/blame.go
+++ b/internal/function/blame.go
@@ -2,6 +2,9 @@ package function
 
 import (
 	"fmt"
+	"io"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/src-d/gitbase"
 	"github.com/src-d/go-mysql-server/sql"
@@ -17,10 +20,11 @@ type BlameGenerator struct {
 	curLine int
 	curFile *object.File
 	lines   []*git.Line
+	ctx     *sql.Context
 }
 
-func NewBlameGenerator(c *object.Commit, f *object.FileIter) (*BlameGenerator, error) {
-	return &BlameGenerator{commit: c, fIter: f, curLine: -1}, nil
+func NewBlameGenerator(c *object.Commit, f *object.FileIter, ctx *sql.Context) (*BlameGenerator, error) {
+	return &BlameGenerator{commit: c, fIter: f, curLine: -1, ctx: ctx}, nil
 }
 
 func (g *BlameGenerator) loadNewFile() error {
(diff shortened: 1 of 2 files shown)


internal/function/blame.go
```diff
@@ -32,7 +36,11 @@ func (g *BlameGenerator) loadNewFile() error {
 
 	result, err := git.Blame(g.commit, g.curFile.Name)
 	if err != nil {
-		return err
+		msg := fmt.Sprintf("Error in BLAME for file %s: %s",
```
erizocosmico: of one of the parameters is in a newline, then better put each in a new line:

```go
msg := fmt.Sprintf(
    "Error in BLAME for file %s: %s",
    g.curFile.Name,
    err.Error(),
)

File: internal/function/blame.go [hunk 2]
```diff
@@ -145,7 +153,7 @@ func (b *Blame) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 		return nil, err
 	}
 
-	bg, err := NewBlameGenerator(commit, fIter)
+	bg, err := NewBlameGenerator(commit, fIter, ctx)
```
erizocosmico: Context is by convention always the first parameter if a function takes a context.

internal/function/blame.go
juanjux: Changed.

internal/function/blame.go
juanjux: Fixed, thanks, learnt something new.

Issue #955, opened 2019-08-23
Non-persistent behavior on logging bblfsh parse errors


lwsanty:
Having [test](https://github.com/src-d/regression-gitbase/blob/9686fdc41c5ec18c0765983362bc02bff3bc2982/cmd/regression-bblfsh-mockups/main.go#L170) in bblfsh-mockups regression suite
One of it's sub-cases uses bblfsh mockup that returns such response
```go
&v2.ParseResponse{Language: "go", Uast: bufData, Errors: nil},
```
where `bufData = getUASTBytes("a")`
```go
func getUASTBytes(text string) ([]byte, error) {
	node := nodes.String(text)
	buf := bytes.NewBuffer(nil)
	if err := nodesproto.WriteTo(buf, node); err != nil {
		return nil, err
	}
	return buf.Bytes(), nil
}
```

**expected** gitbase output
```go
time="2019-08-23T11:44:03+03:00" level=info msg="server started and listening on localhost:3306"
time="2019-08-23T11:44:04+03:00" level=info msg="NewConnection: client 1"
time="2019-08-23T11:44:04+03:00" level=info msg="audit trail" action=authentication address="127.0.0.1:58320" success=true system=audit user=root
time="2019-08-23T11:44:04+03:00" level=info msg="audit trail" action=authorization address="127.0.0.1:58320" connection_id=1 permission=read pid=1 query="select file_path, uast(blob_content) name \nfrom refs natural \njoin commit_files natural \njoin blobs \nwhere LANGUAGE(file_path) = 'Go' \nlimit 10" success=true system=audit user=root
time="2019-08-23T11:44:04+03:00" level=warning msg="unable to parse the given blob using bblfsh: rpc error: code = Canceled desc = context canceled"
time="2019-08-23T11:44:04+03:00" level=info msg="audit trail" action=query address="127.0.0.1:58320" connection_id=1 duration=40.864078ms pid=1 query="select file_path, uast(blob_content) name \nfrom refs natural \njoin commit_files natural \njoin blobs \nwhere LANGUAGE(file_path) = 'Go' \nlimit 10" success=true system=audit user=root
time="2019-08-23T11:44:04+03:00" level=info msg="ConnectionClosed: client 1"
```
the most crucial line to assert is
```go
time="2019-08-23T11:44:04+03:00" level=warning msg="unable to parse the given blob using bblfsh: rpc error: code = Canceled desc = context canceled"
```

**observed** in 20% of this test runs the output does not contain parse error line
[...]

PR #973, opened 2019-10-11
update go-mysql-server version

lwsanty:
required to fix race cases https://github.com/src-d/gitbase/issues/955#issuecomment-535955498
from issue #955

Signed-off-by: lwsanty <lwsanty@gmail.com>


lwsanty:
@agarciamontoro hmm, maybe that's a placebo effect but somehow it fixed data races https://github.com/src-d/gitbase/issues/955#issuecomment-535955498 , I just double-checked it and it passes in 20/20 runs o_0

agarciamontoro:
Talked with @lwsanty over Slack and he could no longer reproduce the bug. We agreed to close this one for now.

